PR #1, opened 2017-05-26
Add Services and Badges

chessbyte:


chessbyte:
Merging and will fix broke test on Ruby 2.4 in a future PR.

Ruby 2.4 collapsed Fixnum and Bignum into Integer (which does not exist in Ruby 2.3).  To resolve the issue, we need to better understand what the purpose of https://github.com/ManageIQ/manageiq-smartstate/blob/master/lib/fs/ntfs/directory_index_node.rb#L55 is

PR #10, opened 2017-06-01
Removed questionable check in NTFS code.

roliveri:
It appears the code is attempting to test for valid data, and raising an error
when the class of the returned data item is Bignum. This should only happen
when the data in question is corrupt, so there shouldn't be an explicit test for it.

Given the test is breaking Ruby 2.4 tests, I'm removing the test.


jerryk55:
Looks good.

PR #13, opened 2017-07-17
Replace File.sizeEx call with File.size

Fryguy:
File.sizeEx is an extension provided by manageiq-gems-pending that is no longer needed.  Its purpose is to handle files larger than 4GB, but File.size handles those now.

@roliveri Please review.

### DIFF
--- a/lib/metadata/VmConfig/VmConfig.rb
+++ b/lib/metadata/VmConfig/VmConfig.rb
@@ -446,7 +446,7 @@ class VmConfig
     @files = []
     if @direct_file_access
       Dir.glob(File.join(@configPath, "/*.*")) do |f|
-        s = File.sizeEx(f) rescue 0
+        s = File.size(f) rescue 0
         @files << {:path => f, :name => File.basename(f), :size => s, :mtime => File.mtime(f)}
       end
     else


Fryguy:
@roliveri I don't understand this error.  Is this coming from my PR or is this failing elsewhere as well?

roliveri:
@Fryguy It looks like the current cassettes are no longer valid, and need to be re-recorded.
However, I'm not sure if this specific change is the cause. Have other Cross-repo changes been made to MiqLargeFile, etc? it looks like the problem is in that area.

Fryguy:
> Have other Cross-repo changes been made to MiqLargeFile, etc? it looks like the problem is in that area.

The only other change was a removal of calls to Win32 specific code, which should not affect anything I would think: https://github.com/ManageIQ/manageiq-gems-pending/pull/239

Fryguy:
@roliveri Master is broken with the same error, so this is unrelated.  How do you want to proceed?

roliveri:
@Fryguy The last PR, #12 was green. So I would think that it had to be a change in another repo that caused this failure. Unless Travis wasn't enabled until after #12 was merged.

The cassettes need to be re-recorded.

PR #16, opened 2017-07-19
Remove dependency on miq-file File.normalize extension

agrare:
File.normalize replaces `\` with `/` and expands relative file paths but
set_vmconfig_path is the only caller.  Move this logic into VmConfig so
we can remove the dependency on the miq-file extension.


agrare:
cc @Fryguy

agrare:
For reference here is the old `File.normalize` method from `gems-pending` https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/extensions/miq-file.rb#L8-L10

agrare:
@roliveri can you review?  The travis failures look unrelated

PR #17, opened 2017-07-28
Fixed test failures.

roliveri:
Re-generated cassette files for miq_disk_spec.


roliveri:
Not sure why the tests are still failing, it's OK when I run them locally. I think it might be because I recorded the cassettes on my MAC - the code path may be different between OS X and Linux.

Fryguy:
Looks like it: https://github.com/ManageIQ/manageiq-smartstate/blob/master/lib/disk/modules/MiqLargeFile.rb#L12-L22

PR #15, opened 2017-07-18
Remove Windows Platform Code

jerryk55:
All references to Windows Platform code have been removed, including
MiqLargeFileWin32 and references to it.
The original issue driving this PR is https://github.com/ManageIQ/manageiq/issues/1404 which was specific to Win32API being removed in Ruby 2.1.  However it seemed a waste of time to not also delete the other two minor places where a test for the Windows platform was made in this repo.

@djberg96 @Fryguy @roliveri please review.  Please merge when appropriate.  Thanks.

PR #18, opened 2017-07-31
Remove code for running on Ruby on Windows

Fryguy:
@roliveri Please review.


lib/disk/modules/MiqLargeFile.rb
```diff
@@ -1,23 +1,17 @@
...
-  end
+if Sys::Platform::IMPL == :linux && MiqSystem.arch == :x86_64
+  require 'linux_block_device'
+  require 'disk/modules/RawBlockIO'
 end
 
 module MiqLargeFile
   def self.open(file_name, flags)
-    case Sys::Platform::OS
-    when :unix
```
Fryguy: Note that `Sys::Platform::OS == :unix` on both MacOS and Linux, but not Windows

Fryguy:
cc @jerryk55 
See also https://github.com/ManageIQ/manageiq-smartstate/pull/15

Fryguy:
Also added some cleanup so that the sys-uname gem is properly required (including creating a method in test_helper so we don't have to require it separately in multiple test files).

chessbyte:
@Fryguy @roliveri we should remove the windows agent support.  just cluttering the code base.

Fryguy:
@roliveri Rebased

PR #19, opened 2017-08-01
Added Rubocop and CodeClimate configuration files

### LINKED ISSUE #14: Needs rubocop files
cc @bdunne @chessbyte

chessbyte:
Fixes #14

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,51 @@
+---
+exclude_paths:
+- ".git/"
+- "**.xml"
+- "**.yaml"
+- "**.yml"
+- "lib/metadata/linux/test/Packages"
+- "lib/metadata/linux/test/tc_LinuxUtils.rb"
+- "locale/"
+- "spec/"
+- "test/"
+- "tools/"
+- "tmp/"
+engines:
+  brakeman:
+    # very slow :sad_panda:
+    enabled: false
+  bundler-audit:
+    # requires Gemfile.lock
+    enabled: false
+  csslint:
+    enabled: false
+  duplication:
+    enabled: true
+    config:
+      languages:
+      - ruby
+      - javascript
+  eslint:
+    enabled: false
+    channel: "eslint-3"
+  fixme:
+    # let's enable later
+    enabled: false
+  markdownlint:
+    # let's enable later
+    enabled: false
+  rubocop:
+    enabled: true
+    config: '.rubocop_cc.yml'
+prepare:
+  fetch:
+  - url: "https://raw.githubusercontent.com/ManageIQ/guides/master/.rubocop_base.yml"
+    path: ".rubocop_base.yml"
+  - url: "https://raw.githubusercontent.com/ManageIQ/guides/master/.rubocop_cc_base.yml"
+    path: ".rubocop_cc_base.yml"
+ratings:
+  paths:
+  - Gemfile.lock
+  - "**.rake"
+  - "**.rb"
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /coverage/
 /doc/
 /pkg/
+/.rubocop-*
 /spec/reports/
 /tmp/
 
--- a/.rubocop_local.yml
+++ b/.rubocop_local.yml
@@ -1,0 +1,8 @@
+#
+# Overrides
+#
+GlobalVars:
+  AllowedVariables:
+  # Loggers
+  - $log
+  - $vim_log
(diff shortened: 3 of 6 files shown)


.rubocop_local.yml
```diff
@@ -0,0 +1,19 @@
...
+  AllowedVariables:
+  # Loggers
+  - $log
+  - $vim_log
+#
+# Special Exclusions
+#
+AllCops:
+  TargetRubyVersion: 2.2
+Metrics/LineLength:
```
bdunne: I feel like we shouldn't add any exclusions for Gemfile in gems since it shouldn't be used.

.rubocop_local.yml
```diff
@@ -0,0 +1,19 @@
...
+  Exclude:
+    - Gemfile
+Style/ExtraSpacing:
```
bdunne: Same here

.rubocop_local.yml
```diff
@@ -0,0 +1,19 @@
...
+GlobalVars:
+  AllowedVariables:
+  # Loggers
+  - $log
+  - $vim_log
+#
+# Special Exclusions
+#
+AllCops:
+  TargetRubyVersion: 2.2
```
bdunne: I feel like this belongs in the rubocop_base.yml on ManageIQ/guides if we need it

.rubocop_local.yml
```diff
@@ -0,0 +1,19 @@
+#
+# Overrides
+#
+GlobalVars:
+  AllowedVariables:
```
bdunne: Should these loggers be allowed?  I think we should encourage manageiq-smartstate to add a logger and log to that instead of continuing to use `$log` and `$vim_log`.  Thoughts?

.codeclimate.yml
```diff
@@ -0,0 +1,46 @@
+---
```
bdunne: I see that there are a few changes from the template file in the generator.  Do you plan to add the generic ones ("**.svg", "**.wsdl", etc.) to the template?

bdunne:
Fixes #14

bdunne:
@chessbyte Can you also add rubocop as a development dependency to the gemspec?

.codeclimate.yml
chessbyte: I just copied https://github.com/ManageIQ/manageiq-gems-pending/blob/master/.codeclimate.yml

.rubocop_local.yml
chessbyte: I just copied https://github.com/ManageIQ/manageiq-gems-pending/blob/master/.rubocop_local.yml

.rubocop_local.yml
chessbyte: I just copied https://github.com/ManageIQ/manageiq-gems-pending/blob/master/.rubocop_local.yml

.rubocop_local.yml
chessbyte: They are still used in this code base, although `$vim_log` is only used in tests.  Someone should remove `$vim_log` in a separate PR.

bdunne:
@chessbyte Can you also add ".rubocop-*" to the .gitignore?

.gitignore
```diff
@@ -5,6 +5,7 @@
 /coverage/
 /doc/
 /pkg/
+/.rubocop*
```
bdunne: This will ignore all `.rubocop` files, we only want to ignore `.rubocop-*`
For example: `.rubocop-https---raw-githubusercontent-com-ManageIQ-guides-master--rubocop-base-yml`

.rubocop_local.yml
chessbyte: @bdunne unsure of the actual suggestion here.  Will follow your guidance on this.

bdunne: I removed it in https://github.com/ManageIQ/manageiq-gems-pending/pull/253 and moved it to guides in https://github.com/ManageIQ/guides/pull/238 so you no longer need it here.

.codeclimate.yml
bdunne: https://github.com/ManageIQ/manageiq/blob/master/lib/generators/provider/templates/.codeclimate.yml is the copy that I would recommend copying from.  I cleaned up the one in manageiq-gems-pending in https://github.com/ManageIQ/manageiq-gems-pending/pull/254

chessbyte:
@bdunne I think I addressed all the issues you raised.  Please re-review and I can squash commits when you give it a 👍

chessbyte:
@Fryguy after multiple reviews with @bdunne, I squashed the commits and re-pushed.  Please review.

PR #21, opened 2017-08-01
Add alternate path to allow SSA of OpenStack instances booted from volumes

tzumainn:
https://bugzilla.redhat.com/show_bug.cgi?id=1341867

### DIFF
--- a/lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
+++ b/lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
@@ -74,8 +74,9 @@ module MiqOpenStackCommon
     while image_service.images.get(image_id).status.downcase != 'active'
       sleep(10)
     end
-    volume.destroy
     return image_service.images.get(image_id)
+  ensure
+    volume.destroy
   end
   
   def get_image_file_glance_v2(image_id)
@@ -89,9 +90,12 @@ module MiqOpenStackCommon
       # try getting image from metadata; oddly the image metadata is only available if
       # image is queried through compute service
       if snapshot_id = get_image_metadata_snapshot_id(compute_service.images.get(image_id))
-        image = create_image_from_snapshot(snapshot_id, image.disk_format)
-        tf = download_image_data_glance_v2(image)
-        image.destroy
+        begin
+          temp_image = create_image_from_snapshot(snapshot_id, image.disk_format)
+          tf = download_image_data_glance_v2(temp_image)
+        ensure
+          temp_image.destroy
+        end
       end
     else
       tf = download_image_data_glance_v2(image)
--- a/lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
+++ b/lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
@@ -63,20 +63,22 @@ module MiqOpenStackCommon
     }
     volume = volume_service.volumes.new(volume_options)
     volume.save
-    while volume.status != 'available'
-      sleep(10)
-      volume = volume_service.volumes.get(volume.id)
+    begin
+      while volume.status != 'available'
+        sleep(10)
+        volume = volume_service.volumes.get(volume.id)
+      end
+      response = volume_service.action(volume.id, 'os-volume_upload_image' => {
+                                         :image_name => "Temp Image from #{snapshot.name}",
+                                         :disk_format => disk_format})
+      image_id = response.body["os-volume_upload_image"]["image_id"]
+      while image_service.images.get(image_id).status.downcase != 'active'
+        sleep(10)
+      end
+      return image_service.images.get(image_id)
+    ensure
+      volume.destroy
     end
-    response = volume_service.action(volume.id, 'os-volume_upload_image' => {
-                                       :image_name => "Temp Image from #{snapshot.name}",
-                                       :disk_format => disk_format})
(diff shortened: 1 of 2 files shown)


tzumainn:
@roliveri I'm not sure the test failures have anything to do with this code, but let me know if I'm missing something?

The codeclimate issue is intentional.

Fryguy:
@tzumainn master was broken but I just merged the PR that will fix it.  I bounced this PR so hopefully it goes green.

roliveri:
@tzumainn 

Let me see if I understand the objects involved in this process:

#### SSA of instance:
1. Snapshot the instance, creating an image (`image1`). If the instance was booted from a Cinder volume (`volume1`), `image1` will contain a pointer to a snapshot of `volume1`.
2. When we attempt to read the data from `image1` via Glance, if the size is zero, we know the image is pointing to a snapshot of `volume1`.
3. So, we then create a new volume, based on the snapshot (`volume2`).
4. We create an image based on `volume2` (`image2`).
5. Destroy `volume2`.
6. Read the data from `image2`.
7. Destroy `image2`.
8. `image1` is destroyed in `delete_evm_snapshot` as it was in the original implementation.
#### SSA of image:
- Steps 2-7 described above.

tzumainn:
@roliveri You're correct! I'll add the ensure block.

tzumainn:
@roliveri I've added the ensure block, thanks!

lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
```diff
@@ -50,6 +53,55 @@ def get_image_file_glance_v2(image_id)
...
+    log_prefix = "#{self.class.name}##{__method__}"
+
+    image = image_service.images.get(image_id)
+    raise "Image #{image_id} not found" unless image
+    $log.debug "#{log_prefix}: image = #{image.class.name}"
+
+    if image.size.to_i == 0
+      # try getting image from metadata; oddly the image metadata is only available if
+      # image is queried through compute service
+      if snapshot_id = get_image_metadata_snapshot_id(compute_service.images.get(image_id))
```
roliveri: @tzumainn
To quiet CodeClimate, could you change the above line to something like:
```ruby
unless (snapshot_id = get_image_metadata_snapshot_id(compute_service.images.get(image_id))).nil?
```

lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
```diff
@@ -50,6 +53,55 @@ def get_image_file_glance_v2(image_id)
...
+    response = volume_service.action(volume.id, 'os-volume_upload_image' => {
+                                       :image_name => "Temp Image from #{snapshot.name}",
+                                       :disk_format => disk_format})
+    image_id = response.body["os-volume_upload_image"]["image_id"]
+    while image_service.images.get(image_id).status.downcase != 'active'
+      sleep(10)
+    end
+    return image_service.images.get(image_id)
+  ensure
+    volume.destroy
```
roliveri: If we fail before the line:
```ruby
volume = volume_service.volumes.new(volume_options)
```
then `volume` won't be set when we enter the ensure block, causing a runtime error.
The `ensure` block should be associated with a `begin` block that starts after the `volume =` line.

lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
```diff
@@ -50,6 +53,55 @@ def get_image_file_glance_v2(image_id)
...
+    image = image_service.images.get(image_id)
+    raise "Image #{image_id} not found" unless image
+    $log.debug "#{log_prefix}: image = #{image.class.name}"
+
+    if image.size.to_i == 0
+      # try getting image from metadata; oddly the image metadata is only available if
+      # image is queried through compute service
+      if snapshot_id = get_image_metadata_snapshot_id(compute_service.images.get(image_id))
+        begin
+          temp_image = create_image_from_snapshot(snapshot_id, image.disk_format)
```
roliveri: @tzumainn 
The `begin` block should start after the `temp_image =` line, because if `create_image_from_snapshot` fails, `temp_image` won't be set in the `ensure` block.

tzumainn:
@roliveri Thanks again for the review! I've fixed the issues you pointed out, let me know if you find anything else!

lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
```diff
@@ -50,6 +53,57 @@ def get_image_file_glance_v2(image_id)
...
+  
+  def create_image_from_snapshot(snapshot_id, disk_format)
+    snapshot = volume_service.snapshots.get(snapshot_id)
+    volume_options = {
+      :name        => "Temp Volume from #{snapshot.name}",
+      :size        => snapshot.size,
+      :snapshot_id => snapshot_id
+    }
+    volume = volume_service.volumes.new(volume_options)
+    volume.save
```
roliveri: @tzumainn 
If the `volume.save` raises an error, does it mean the volume doesn't need to be destroyed?
If not, then the `volume.save` should be moved inside the `begin` block.

tzumainn: Correct - the volume shouldn't persist unless volume.save succeeds.

PR #23, opened 2017-08-21
Handle Managed Disk Snapshots

jerryk55:
If the Azure VM has a Managed Disk then use the new Managed Disk
disk module to read its snapshot.

On this go-round the snapshot will only be required for Managed Disks.

This is one of several PRs required to add support for SSA for Managed Disks,
that is is response to the following BZs:
https://bugzilla.redhat.com/show_bug.cgi?id=1475540
and
https://bugzilla.redhat.com/show_bug.cgi?id=1459612

A PR for the azure-armrest gem has already been merged in advance of this:
https://github.com/ManageIQ/azure-armrest/pull/299

A PR for the ManageIQ repo has been added as well to work with this - 
https://github.com/ManageIQ/manageiq/pull/15865

A PR for the manageiq-providers-azure repo has also been added.
https://github.com/ManageIQ/manageiq-providers-azure/pull/117

Links
----------------

* https://bugzilla.redhat.com/show_bug.cgi?id=1475540
* https://bugzilla.redhat.com/show_bug.cgi?id=1459612
* https://github.com/ManageIQ/azure-armrest/pull/299
* https://github.com/ManageIQ/manageiq/pull/15865
* https://github.com/ManageIQ/manageiq-providers-azure/pull/117

Steps for Testing/QA
-------------------------------

Add an Azure VM with a Managed Disk.
Run SmartState Analysis against the VM.

@roliveri @hsong-rh please review.

### DIFF
--- a/lib/disk/modules/AzureCommon.rb
+++ b/lib/disk/modules/AzureCommon.rb
@@ -1,0 +1,95 @@
+require_relative '../MiqDisk'
+require 'ostruct'
+
+module AzureCommon
+  # The maximum read length that supports MD5 return.
+  MAX_READ_LEN  = 1024 * 1024 * 4
+  SECTOR_LENGTH = 512
+
+  def self.d_init_common(dInfo)
+    @blockSize        = SECTOR_LENGTH
+    @blob_uri         = dInfo.blob_uri if dInfo.blob_uri
+    @disk_name        = dInfo.disk_name if dInfo.disk_name
+    @storage_acct_svc = dInfo.storage_acct_svc if dInfo.storage_acct_svc
+    @storage_disk_svc = dInfo.storage_disk_svc if dInfo.storage_disk_svc
+    @resource_group   = dInfo.resource_group if dInfo.resource_group
+
+    if @storage_acct_svc
+      @my_class     = "AzureBlobDisk"
+      uri_info      = @storage_acct_svc.parse_uri(@blob_uri)
+      @container    = uri_info[:container]
+      @blob         = uri_info[:blob]
+      @acct_name    = uri_info[:account_name]
+      @snapshot     = uri_info[:snapshot]
+      @storage_acct = @storage_acct_svc.accounts_by_name[@acct_name]
+      @disk_path    = @blob_uri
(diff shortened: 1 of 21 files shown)


jerryk55:
@roliveri Yes the non-managed disk snapshot support will be delivered in a subsequent PR after this batch is merged.  And I will look at the codeclimate issues.

lib/disk/modules/AzureCommon.rb
```diff
@@ -0,0 +1,95 @@
+require_relative '../MiqDisk'
+require 'ostruct'
+
+module AzureCommon
```
roliveri: @jerryk55 I think this name isn't specific enough. Would `AzureDiskCommon` be better?

jerryk55: @@roliveri good idea.  BTW I still have some code climate issues to fix up, and for some reason the math in the spec tests is off.  I will work with that in the morning.

jerryk55:
@roliveri the failure above is intermittent both in Travis and in my appliance when run from the command line.  Trying to discern the issue but it is perplexing.  Originally CodeClimate complained about 9 issues - its down to 2 now with the issue being slightly duplicated code in the new methods of the two disk modules.    I am currently at a loss for the spec test failures.  Any suggestions?  I will also need to add AzureManagedDisk tests before this should be merged.

roliveri:
@jerryk55 It's strange. The numbers being read from the recorded cassettes don't match the expected values. So, unless the cassettes were re-recorded, how could the numbers change? You have to check the code that reads the info from armrest and see if it manipulates the values before returning, (lba, etc).
Maybe that code changed, but then I would think the tests would have failed when the PR for that change was made.

jerryk55:
Ok @roliveri problem with the tests resolved.  Still need Managed Disk tests.

jerryk55:
@roliveri I believe this can be merged now.  New tests added and passing.

jerryk55:
Attempting to restart CodeClimate checks.

jerryk55:
@miq-bot add_label fine/yes

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq-gems-pending/pull/267

PR #26, opened 2017-09-09
Snapshot Support for Non-Managed Disks

jerryk55:
In the case of non-Managed (Blob) disks, append the snapshot
name (which is in all actuality the date tag returned from
azure-armrest for the snapshot) to the URI to be used to read
the disk.

This is one of several PRs required to fix https://bugzilla.redhat.com/show_bug.cgi?id=1463780.
The PR against ManageIQ/manageiq is https://github.com/ManageIQ/manageiq/pull/15960, and one against ManageIQ/manageiq-providers-azure will be noted here when it is added.

This PR can be merged in any order since it checks for a tag to be passed in and only uses that snapshot tag if it is there.

@roliveri @hsong-rh please review.  I'm not sure if we need to back port this to Fine but it wouldn't be a bad idea.

### DIFF
--- a/lib/MiqVm/miq_azure_vm.rb
+++ b/lib/MiqVm/miq_azure_vm.rb
@@ -15,7 +15,7 @@ class MiqAzureVm < MiqVm
     if args[:image_uri]
       @uri = args[:image_uri]
     elsif args[:resource_group] && args[:name]
-      vm_obj = vm_svc.get(@name, args[:resource_group])
+      vm_obj = vm_svc.get(@name, @resource_group.name)
       os_disk = vm_obj.properties.storage_profile.os_disk
       if vm_obj.managed_disk?
         #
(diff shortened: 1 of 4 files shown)


lib/MiqVm/miq_azure_vm.rb
```diff
@@ -24,9 +24,10 @@ def initialize(azure_handle, args)
         @snap_name = os_disk.name + "__EVM__SSA__SNAPSHOT"
       else
         #
-        # Non-Managed Disk Snapshot handling will be added here by a separate PR.
+        # Non-Managed Disk Snapshot handling
         #
         @uri = os_disk.vhd.uri
+        @uri << "?snapshot=" << args[:snapshot] if args[:snapshot]
```
roliveri: Is
```ruby
@uri << "?snapshot=#{args[:snapshot]}" if args[:snapshot]
```
preferred?

jerryk55: Sure makes sense.

jerryk55:
@miq-bot add_label fine/yes

simaishi:

```
$ git log -1
commit 3ff4035f4b8da8cd5798bd598587c46eb0d95484
Date:   Mon Sep 11 16:05:31 2017 -0400

    Merge pull request #26 from jerryk55/snapshot_nonmanaged_disks
    
    Snapshot Support for Non-Managed Disks
    (cherry picked from commit 0c525af5d9a6f478193b182d3f5932df6c1a59e2)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1491310
```

PR #28, opened 2017-09-19
Handle case where attempted LVM2 Thin device reads fail

movitto:
As it stands if a client invokes a disk read against a LVM2 Thin Provisioned
volume for which space has not yet been allocated for the address being read,
a rather confusing / ugly error will occur:

  LVM2Thin cannot find device block: N (closest: X)]

  Where N is the data block corresponding to the address being looked up
  and X is the last allocated data block

This patch catches this case and simply returns an empty string if the read
cannot be completed

### DIFF
--- a/lib/VolumeManager/LVM/lvm2disk.rb
+++ b/lib/VolumeManager/LVM/lvm2disk.rb
@@ -43,10 +43,14 @@ module Lvm2DiskIO
       device_id = logicalVolume.thin_segment.device_id
       thin_pool = logicalVolume.thin_pool_volume
       data_blks = thin_pool.metadata_volume.superblock.device_to_data(device_id, pos, len)
+
       data_blks.each do |offset, len|
         thin_pool.data_volume.disk.seek offset
         retStr << thin_pool.data_volume.disk.read(len)
       end
+
+      # add zeros to result if reading from unallocated memory
+      retStr << thin_pool.metadata_volume.superblock.device_fill_data(device_id, pos, len)
 
       return retStr
     end
--- a/lib/VolumeManager/LVM/lvm2disk.rb
+++ b/lib/VolumeManager/LVM/lvm2disk.rb
@@ -44,14 +44,13 @@ module Lvm2DiskIO
       thin_pool = logicalVolume.thin_pool_volume
       data_blks = thin_pool.metadata_volume.superblock.device_to_data(device_id, pos, len)
 
-      data_blks.each do |device_blk, data_blk, blk_offset, blk_len|
+      data_blks.each do |_device_blk, data_blk, blk_offset, blk_len|
         if data_blk.nil?
           # fill in unallactored data
-          puts blk_len
           retStr << Array.new(blk_len, 0).pack("C*")
 
         else
-          thin_pool.data_volume.disk.seek blk_offset
+          thin_pool.data_volume.disk.seek(blk_offset)
           retStr << thin_pool.data_volume.disk.read(blk_len)
         end
       end
--- a/lib/VolumeManager/LVM/lvm2disk.rb
+++ b/lib/VolumeManager/LVM/lvm2disk.rb
@@ -44,13 +44,17 @@ module Lvm2DiskIO
       thin_pool = logicalVolume.thin_pool_volume
       data_blks = thin_pool.metadata_volume.superblock.device_to_data(device_id, pos, len)
 
-      data_blks.each do |offset, len|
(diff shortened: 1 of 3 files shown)


movitto:
@roliveri please review

lib/VolumeManager/LVM/lvm2disk.rb
```diff
@@ -42,7 +42,12 @@ def d_read(pos, len)
     if logicalVolume.thin?
       device_id = logicalVolume.thin_segment.device_id
       thin_pool = logicalVolume.thin_pool_volume
-      data_blks = thin_pool.metadata_volume.superblock.device_to_data(device_id, pos, len)
+      begin
+        data_blks = thin_pool.metadata_volume.superblock.device_to_data(device_id, pos, len)
+      rescue
+        return retStr
```
roliveri: @movitto Should it return an empty string, or a string of 0x0 of the requested length?
That is, unless the length causes the read to exceed the end of the device, then it should return a string of 0x0
of the available length.

Reads from a block device should only return an empty string when at the end of the device. I would think LVM2 would maintain that behavior.

movitto:
@roliveri addressed your feedback. Now the lvm2thin module will fill in the remaining length with zero's. So for example if the invoker tries to read 512 bytes starting at 256 bytes from the end of the allocated space they will get the last 256 bytes of data plus 256 0's

As it stands, the lvm2thin d_read method will return zeros if the user requests a read beyond the total lvm volume size but I believe this edge case is handled elsewhere correct? (in the general case where the user tries to read beyond the extent of a disk boundary)

roliveri:
@movitto Right, I believe the `MiqDisk` will check the boundary  of the underlying device.

roliveri:
@movitto What if the unallocated space falls within the extents of a read?
It is possible to read say, 3 blocks, and the second block is unallocated.

So, we need to check for unallocated data on a per-block basis, filling with zeros
as needed. Unallocated data can be encountered at the beginning, middle or end of a read.

movitto:
@roliveri updated, the lvm2thin module now return zeros for all unallocated blocks

lib/VolumeManager/LVM/lvm2disk.rb
```diff
@@ -43,9 +43,17 @@ def d_read(pos, len)
...
       thin_pool = logicalVolume.thin_pool_volume
       data_blks = thin_pool.metadata_volume.superblock.device_to_data(device_id, pos, len)
-      data_blks.each do |offset, len|
-        thin_pool.data_volume.disk.seek offset
-        retStr << thin_pool.data_volume.disk.read(len)
+
+      data_blks.each do |device_blk, data_blk, blk_offset, blk_len|
+        if data_blk.nil?
+          # fill in unallactored data
+          puts blk_len
```
roliveri: Remove the `puts`.

lib/VolumeManager/LVM/lvm2disk.rb
```diff
@@ -43,9 +43,17 @@ def d_read(pos, len)
       device_id = logicalVolume.thin_segment.device_id
       thin_pool = logicalVolume.thin_pool_volume
       data_blks = thin_pool.metadata_volume.superblock.device_to_data(device_id, pos, len)
-      data_blks.each do |offset, len|
-        thin_pool.data_volume.disk.seek offset
-        retStr << thin_pool.data_volume.disk.read(len)
+
+      data_blks.each do |device_blk, data_blk, blk_offset, blk_len|
```
roliveri: Change `device_blk` to `_device_blk` to quiet code climate.

movitto:
@roliveri updated,

simaishi:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1488378

@movitto  Is it ok to backport this to manageiq-gems-pending Fine branch?

movitto:
@simaishi hey Satoe, this one go lost somewhere in the gaps. I don't see any reason why this fix could not be backported to previous releases. Go for it.

simaishi:

```
$ git log -1
commit fa86b88c418c00fece26db84ccffc823068cf81c
Date:   Fri Sep 22 11:44:43 2017 -0400

    Merge pull request #28 from movitto/preallocated_lvm2thin_fix
    
    Handle case where attempted LVM2 Thin device reads fail
    (cherry picked from commit 9ae19959eee351a71919996464bca31bea45c336)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1511147
```

simaishi:
@miq-bot add_label fine/backported

PR #32, opened 2017-10-12
MiqVm doesn't need to be executable

agrare:
Mark MiqVm.rb as non-executable


agrare:
cc @roliveri @Fryguy 

@roliveri I'm assuming we don't need to execute this file but can you confirm :)

PR #34, opened 2017-10-13
Fix require since miq-file extension is removed

Fryguy:
@roliveri @agrare Please review.

I missed this because the file doesn't use any of the miq-file methods, and I didn't catch it when grepping for the removed methods.  Looks like we should have removed it when we removed File.normalize.


Fryguy:
Ref: https://github.com/ManageIQ/manageiq-gems-pending/pull/284

PR #35, opened 2017-10-13
Replace old File.normalize with expand_path and tr

agrare:
File.normalize was removed in https://github.com/ManageIQ/manageiq-gems-pending/pull/245
but this caller was missed, replace it with a simple expand_path and
tr('\\', '/').


agrare:
Should be green once https://github.com/ManageIQ/manageiq-smartstate/pull/34 is merged

agrare:
@miq-bot add-label bug

PR #36, opened 2017-10-16
Remove path_to_uri from smartstate util

agrare:
The path_to_uri method is only used by the SCVMM refresh parser so we
can move this method there.

Depends on: https://github.com/ManageIQ/manageiq-providers-scvmm/pull/22
cc @Fryguy


Fryguy:
Be sure to also kill the addressable gem (and I don't think you moved the specs over to scvmm either).

agrare:
Okay moved addressable and the specs over to scvmm here https://github.com/ManageIQ/manageiq-providers-scvmm/pull/25

PR #37, opened 2017-10-17
Changes a non-fatal error message to be reported info log level

movitto:
Missing parent directories in EXT4 getfile module will no longer report
an error (inappropriate when checking for file existance, etc)

### DIFF
--- a/lib/fs/MiqFS/modules/Ext4.rb
+++ b/lib/fs/MiqFS/modules/Ext4.rb
@@ -5,6 +5,8 @@ require 'fs/ext4/superblock'
 require 'fs/ext4/directory_entry'
 require 'fs/ext4/directory'
 
+require 'fs/MiqFsError'
+
 # Ext4 file system interface to MiqFS.
 module Ext4
   # Default directory cache size.
@@ -225,6 +227,10 @@ module Ext4
     begin
       dirObj = ifs_getDir(dir, miqfs)
       dirEnt = dirObj.nil? ? nil : dirObj.findEntry(fname)
+
+    rescue DirectoryNotFound => err
+      dirEnt = nil
+
     rescue RuntimeError => err
       $log.error err.message.to_s if $log
       dirEnt = nil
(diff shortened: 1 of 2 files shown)


movitto:
@roliveri please review, this fixes BZ #1450008 (though we may wish to incorporate a different fix as mentioned in the BZ)

lib/fs/MiqFS/modules/Ext4.rb
```diff
@@ -226,7 +226,7 @@ def ifs_getFile(p, miqfs = nil)
       dirObj = ifs_getDir(dir, miqfs)
       dirEnt = dirObj.nil? ? nil : dirObj.findEntry(fname)
     rescue RuntimeError => err
-      $log.error err.message.to_s if $log
+      $log.info err.message.to_s if $log
```
Fryguy: What about other RuntimeErrors?  I think this should be more prescriptive either by catching a specific error or by inspecting the log message.

movitto:
@roliveri updated the PR to remove the log message & bring it in line w/ the other FS modules.

@Fryguy I'm not fully sure I understand, can you elaborate? As it stands, if there are any problems or errors during the file / directory lookup, this will catch them and return nil. While it may sense to log missing files elsewhere, it doesn't make sense to include this ala-carte log message during the File#exists? operation (for example) which dispatches to this method.

movitto:
@roliveri @Fryguy just updated this to define a new Directory Not Found Error Class (subclass of runtime error) and use in this operation. We can expand usage to the other modules after this one is merged.

lib/fs/MiqFsError.rb
```diff
@@ -0,0 +1,11 @@
+class MiqFsError < RuntimeError
+end
+
+class DirectoryNotFound < MiqFsError
```
roliveri: @movitto 
I'm concerned about having a generic class name like this defined in the global namespace. Normally, I'd suggest we define these classes under a namespace like `MiqFs::DirectoryNotFound` and maybe even `MiqFS::Error` as well. But since most of this will be changing going forward, maybe we should just make the name less generic, like `MiqFsDirectoryNotFound`.

movitto:
@roliveri changed the name

PR #41, opened 2017-10-31
Fix Resource Group Object Issue with Smartstate on Azure Blob Disk Instances

jerryk55:
Azure Blob Disk instances have an issue caused by a previous fix for Managed Images
back-ported to Fine.  This issue is only in Gaprindashvili and Master.
An older modification to the Azure provider changed the ResourceGroup
from a string to an object and this is a result of that change.

This fixes BZ https://bugzilla.redhat.com/show_bug.cgi?id=1507977.
This should be back-ported to Gaprindashvili but NOT to Fine, which does not have the issue.
In order to do this we need to make sure that the Gemfile for manageiq-providers-azure contains the correct version of the manageiq-smartstate gem after this change is pushed out.

@roliveri @hsong-rh please review and merge.

### DIFF
--- a/lib/MiqVm/miq_azure_vm.rb
+++ b/lib/MiqVm/miq_azure_vm.rb
@@ -8,7 +8,7 @@ class MiqAzureVm < MiqVm
     @azure_handle   = azure_handle
     @uri            = nil
     @disk_name      = nil
-    @resource_group = args[:resource_group]
+    @resource_group = args[:resource_group]&.name
     @managed_image  = args[:managed_image]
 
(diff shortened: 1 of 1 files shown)


jerryk55:
@miq-bot add_label gaprindashvili/yes

jerryk55:
@miq-bot add_label fine/no

jerryk55:
@miq-bot add_label "smart state"

jerryk55:
I am going to close this PR and move the relevant fix into the Azure provider repo instead.  The provider should know whether the Resource Group is a string or an Object and always pass the string to the manageiq-smartstate gem.

PR #42, opened 2017-11-01
Force Bundler 1.15.4 on CI

chrisarcand:
There appears to be a regression of some sort with the new bundler
version, causing azure-armrest to be pulled in as an ancient 0.0.3
version which causes builds to fail. Pulling this back to the last known
good version of Bundler before investigating.

cc/ @blomquisg @jrafanie

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,9 @@ language: ruby
 rvm:
   - 2.3.1
   - 2.4.2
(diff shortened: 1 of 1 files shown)


chrisarcand:
https://github.com/bundler/bundler/issues/6139

Fryguy:
:+1:  Great find!

PR #43, opened 2017-11-01
Use latest Bundler (1.16.0+)

chrisarcand:
A regression was found in Bundler 1.16.0 that caused us to force an old version; the problem disappears with specifying a minimum dependency requirement on azure-armrest which arguably we should be doing anyway.  Thus, let's do that and go back to using what Travis gives us and not keep this baggage.

Links:
* https://github.com/ManageIQ/manageiq-smartstate/pull/42
* https://github.com/bundler/bundler/issues/6139

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ language: ruby
 rvm:
   - 2.3.1
   - 2.4.2
-before_install:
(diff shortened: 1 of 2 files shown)


chrisarcand:
@miq-bot assign @Fryguy

PR #44, opened 2017-11-02
Use the Smartstate Snapshot Name Built by the Provider

jerryk55:
Since there is an 80 character limit on the length of the snapshot name
constructed for managed disks use the name set by the Azure provider.

This is one of two PRs required to fix https://bugzilla.redhat.com/show_bug.cgi?id=1508577.
The other will be against the Azure provider repo.

This change is required upstream, for Fine, and for Gaprindashvili.  Once this gem is
pushed out the Gemfile for the Azure provider will be updated in all three places.

@roliveri @hsong-rh please review and merge.  This PR should be merged asap since the
The Azure Provider PR is https://github.com/ManageIQ/manageiq-providers-azure/pull/157.
has already been merged.

### DIFF
--- a/lib/MiqVm/miq_azure_vm.rb
+++ b/lib/MiqVm/miq_azure_vm.rb
@@ -23,15 +23,8 @@ class MiqAzureVm < MiqVm
       if vm_obj.managed_disk?
         #
         # Use the Smartstate SNAPSHOT Added by the Provider
-        # If the provider didn't pass the name in then it is running older code that built
-        # the old name - use that instead
         #
-        if args[:snapshot]
-          @disk_name = args[:snapshot]
-        else
-          @disk_name = os_disk.name + "__EVM__SSA__SNAPSHOT"
-          $log.warn("MiqAzureVm: missing required snapshot arg: using old name #{@disk_name}")
-        end
+        raise ArgumentError, "MiqAzureVm: missing required arg :snapshot for Managed Disk" unless (@disk_name = args[:snapshot])
       else
         #
         # Non-Managed Disk Snapshot handling
--- a/lib/MiqVm/miq_azure_vm.rb
+++ b/lib/MiqVm/miq_azure_vm.rb
@@ -22,9 +22,16 @@ class MiqAzureVm < MiqVm
       os_disk = vm_obj.properties.storage_profile.os_disk
       if vm_obj.managed_disk?
         #
-        # Use the EVM SNAPSHOT Added by the Provider
+        # Use the Smartstate SNAPSHOT Added by the Provider
+        # If the provider didn't pass the name in then it is running older code that built
+        # the old name - use that instead
         #
-        @disk_name = os_disk.name + "__EVM__SSA__SNAPSHOT"
+        if args[:snapshot]
+          @disk_name = args[:snapshot]
+        else
+          @disk_name = os_disk.name + "__EVM__SSA__SNAPSHOT"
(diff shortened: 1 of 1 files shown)


lib/MiqVm/miq_azure_vm.rb
```diff
@@ -22,9 +22,16 @@ def initialize(azure_handle, args)
...
-        # Use the EVM SNAPSHOT Added by the Provider
+        # Use the Smartstate SNAPSHOT Added by the Provider
+        # If the provider didn't pass the name in then it is running older code that built
+        # the old name - use that instead
         #
-        @disk_name = os_disk.name + "__EVM__SSA__SNAPSHOT"
+        if args[:snapshot]
+          @disk_name = args[:snapshot]
+        else
+          @disk_name = os_disk.name + "__EVM__SSA__SNAPSHOT"
```
roliveri: @jerryk55 Under what conditions will we enter this `else` block? I don't understand when this code will be paired with older provider code.

If we can enter this `else` block, isn't it the same problem we had in the first place?

jerryk55: @roliveri I was concerned that there could be a situation where the old provider code was running with the new Smartstate gem and if so then the name of the snapshot would need to be the same as it was previous to this update.  (more on that below)

If we enter this block and the old name is being used, then yes we have an issue if the name is too long - but only if the name is too long - and then we are no worse off than we were before.  

If we aren't passing in the name of the snapshot as an argument then the provider already snapshotted the disk using the old name - if we don't use that name then *any* attempt at running SSA here will fail. If we do use the old name then it will only fail if it is too long.

The above all being said, now that the provider PR has been merged *before* this PR,  this line of execution is actually moot and can probably be removed.  I'll take it out....

jerryk55:
@roliveri the code to allow the old-style snapshot name has been removed and it will not be an exception if the snapshot name is not passed in as an argument.

simaishi:
@jerryk55 I assume this should be `fine/yes` as https://github.com/ManageIQ/manageiq-providers-azure/pull/157 is `fine/yes`?

jerryk55:
@simaishi yes except that this gem can't be fine/yes since its a gem that gets published.

simaishi:
@jerryk55 You can still mark as `fine/yes` as I know to backport to manageiq-gems-pending 😄

jerryk55:
@miq-bot add_label fine/yes

jerryk55:
@miq-bot add_label gaprindashvili/yes

jerryk55:
@simaishi it let me apply fine/yes but not gaprindashvili/yes

simaishi:
Gaprindashvili uses the released gem, so backport of this PR is needed only to Fine branch. The PR that bumps manageiq-smartstate gem version needs to be marked as `gaprindashvili/yes` instead.

simaishi:

```
$ git log -1
commit 9a76effbec67c302b852917251594237c4b3ad25
Date:   Mon Nov 6 11:45:14 2017 -0500

    Merge pull request #44 from jerryk55/diskname_longer_than_60chars_fails_snap
    
    Use the Smartstate Snapshot Name Built by the Provider
    (cherry picked from commit 02ebbc23a57ef8283c113199847cf4092c66466a)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1512967
```

simaishi:
@miq-bot add_label fine/backported

simaishi:
@miq-bot remove_label fine/yes

PR #50, opened 2018-01-19
Managed Disk Read Semantics Change

jerryk55:
Azure Managed Disk Instances have been taking exceptionally long in
some locations. It turns out that the semantic that azure-armrest
uses to perform reads via the get_blob_raw method is the culprit.
The method gets a SAS URL - an access token to the disk - each
time it performs a read operation. The call to get the access token
is asynchronous and if it does not return immediately with success when
poll is called the gem sleeps for 5 seconds. This adds 5 seconds to
every call to get data.

What the disk module will now do:

Call open on the service (disk service or snapshot service).
Open returns a ManagedDisk object.
Call read on the ManagedDisk object as many times as you need, then call
close on the ManagedDisk object when finished.

Under the covers the ManagedDisk will contain the SAS URL to be used.
In the case that a read on the ManagedDisk returns a Forbidden error,
the assumption is that the SAS URL is no longer any good (perhaps it has
timed out) and we will get a new ManagedDisk object and continue.

Depends on PR github.com/ManageIQ/azure-armrest/pull/355

@blomquisg @djberg96 @hsong-rh @roliveri for your reading and reviewing pleasure.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1508154

### DIFF
--- a/lib/disk/modules/AzureDiskCommon.rb
+++ b/lib/disk/modules/AzureDiskCommon.rb
@@ -43,6 +43,7 @@ module AzureDiskCommon
   end
 
   def d_close_common
+    @managed_disk&.close
     return nil unless $log.debug?
     t1 = Time.now.to_i
     $log.debug("#{@my_class}: close(#{@disk_path})")
@@ -84,7 +85,16 @@ module AzureDiskCommon
       options[:date] = @snapshot if @snapshot
       ret = @storage_acct.get_blob_raw(@container, @blob, key, options)
     else
-      ret = @storage_disk_svc.get_blob_raw(@disk_name, @resource_group, options)
+      retries = 0
+      begin
+        ret = managed_disk.read(options)
+      rescue Azure::Armrest::ForbiddenException => err
+        raise err if retries > 0
+        $log.warn("#{@class}: blob_read: #{err}: acquiring new SAS URL")
+        @managed_disk = nil
+        retries += 1
+        retry
+      end
     end
 
     @reads += 1
@@ -95,5 +105,9 @@ module AzureDiskCommon
 
   def key
     @key ||= @storage_acct_svc.list_account_keys(@storage_acct.name, @storage_acct.resource_group).fetch('key1')
+  end
+
+  def managed_disk
+    @managed_disk ||= @storage_disk_svc.open(@disk_name, @resource_group)
   end
 end
--- a/lib/disk/modules/AzureDiskCommon.rb
+++ b/lib/disk/modules/AzureDiskCommon.rb
@@ -85,16 +85,7 @@ module AzureDiskCommon
       options[:date] = @snapshot if @snapshot
       ret = @storage_acct.get_blob_raw(@container, @blob, key, options)
     else
-      retries = 0
-      begin
-        ret = managed_disk.read(options)
-      rescue Azure::Armrest::ForbiddenException => err
-        raise err if retries > 0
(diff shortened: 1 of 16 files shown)


lib/disk/modules/AzureDiskCommon.rb
```diff
@@ -84,7 +85,16 @@ def blob_read(start_byte, length)
       options[:date] = @snapshot if @snapshot
       ret = @storage_acct.get_blob_raw(@container, @blob, key, options)
     else
-      ret = @storage_disk_svc.get_blob_raw(@disk_name, @resource_group, options)
+      retries = 0
+      begin
+        ret = managed_disk.read(options)
+      rescue Azure::Armrest::ForbiddenException => err
+        raise err if retries > 0
+        $log.warn("#{@class}: blob_read: #{err}: acquiring new SAS URL")
```
roliveri: Shouldn't this retry logic be in the `ManagedDisk` object itself? I thought that was one of the main reasons for implementing that class. The clients should be insulated from all of this.

jerryk55: Let me see if I can do that given the restrictions I found myself working under in the disk_service / snapshot_service classes and the included modules.

jerryk55:
@roliveri ok I've modified this PR and am modifying the dependent azure-armrest PR to do the same.

jerryk55:
The above test will fail until the dependent azure-armrest PR is merged.

jerryk55:
depends_on https://github.com/ManageIQ/azure-armrest/pull/355

jerryk55:
@roliveri tests are now passing with re-recorded cassettes.  This is ready to rock and roll.

roliveri:
@jerryk55 Where should this be backported?

jerryk55:
Probably Fine and Gaprindashvilli.

roliveri:
@jerryk55 There is no Gaprindashvilli/yes label available.

jerryk55:
@roliveri I know.  It ends up going to Fine by way of manageiq-gems-pending.  I think it just gets picked up in Gaprindashvili by way of a gemspec/Gemfile if necessary.

simaishi:

```
$ git log -1
commit f388746aa8bf7f9b3386f8f0f1a498b5d1b767d0
Date:   Tue Jan 23 15:34:19 2018 -0500

    Merge pull request #50 from jerryk55/azure-armrest-speed-refactoring
    
    Managed Disk Read Semantics Change
    (cherry picked from commit 9f859ccdd472980f58938aa7614989f7cb64a0af)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1541163
```

simaishi:
@miq-bot add_label fine/backported,blocker
@miq-bot remove_label fine/yes

simaishi:
@jerryk55 Travis is failing in Fine now...

```
/home/travis/build/ManageIQ/manageiq-gems-pending/lib/gems/pending/disk/modules/AzureDiskCommon.rb:46: syntax error, unexpected '.' (SyntaxError)
    @managed_disk&.close
                   ^
```
https://travis-ci.org/ManageIQ/manageiq-gems-pending/jobs/336312835

Looks to be the issue with unsupported `&.` operation in this version? Can you please create a PR to fix this in gems-pending repo?

jerryk55:
@simaishi I will do so in the AM.

jerryk55:
https://github.com/ManageIQ/manageiq-gems-pending/pull/335

PR #57, opened 2018-03-02
Update to Azure-Armrest Gem Version 0.9.6

jerryk55:
Upgrade manageiq-smartstate to the latest version of the azure-armrest gem.

@roliveri @hsong-rh @djberg96 please review.


djberg96:
👍

PR #61, opened 2018-05-04
Code Climate and RuboCop Errors

jerryk55:
Work on Code Climate Cleanup for the manageiq-smartstate repo.
This commit does not contain fixes for any of the complexity issues
identified by Code Climate as of this point.
@roliveri @hsong-rh if you can being preliminary review I would appreciate it.  Thanks.

### DIFF
--- a/lib/fs/MiqFsUtil.rb
+++ b/lib/fs/MiqFsUtil.rb
@@ -208,8 +208,8 @@ class MiqFsUtil
           toFile = File.join(destDir, i)
           makePath(File.dirname(toFile))
 
-          check_encrypt(cs, i, toFile)
-          check_compress(cs, i, toFile)
+          next if check_encrypt(cs, i, toFile)
+          next if check_compress(cs, i, toFile)
           copyFile(i, toFile)
           next
         end
@@ -233,8 +233,8 @@ class MiqFsUtil
           toFile = File.join(destDir, path)
           makePath(File.dirname(toFile))
 
-          check_encrypt(cs, i, toFile)
-          check_compress(cs, i, toFile)
+          next if check_encrypt(cs, i, toFile)
+          next if check_compress(cs, i, toFile)
 
           copyFile(path, toFile)
         end
@@ -294,9 +294,10 @@ class MiqFsUtil
     if cs.encrypt && cs.encrypt.detect { |e| i =~ e }
       if !cs.noencrypt || !cs.noencrypt.detect { |ne| i =~ ne }
         compressFile(i, toFile)
-        next
+        return true
       end
     end
+    false
   end
 
   def check_compress(cs, i, toFile)
@@ -308,9 +309,10 @@ class MiqFsUtil
     if cs.compress && cs.compress.detect { |e| i =~ e }
       if !cs.nocompress || !cs.nocompress.detect { |ne| i =~ ne }
         compressFile(i, toFile)
-        next
+        return true
       end
     end
+    false
   end
 
   def makePath(path)
--- a/lib/fs/MiqFsUtil.rb
+++ b/lib/fs/MiqFsUtil.rb
@@ -208,7 +208,8 @@ class MiqFsUtil
           toFile = File.join(destDir, i)
           makePath(File.dirname(toFile))
 
-          checkCompress(cs, i, toFile)
+          check_encrypt(cs, i, toFile)
+          check_compress(cs, i, toFile)
           copyFile(i, toFile)
           next
         end
@@ -232,7 +233,9 @@ class MiqFsUtil
           toFile = File.join(destDir, path)
           makePath(File.dirname(toFile))
 
-          checkCompress(cs, i, toFile)
+          check_encrypt(cs, i, toFile)
+          check_compress(cs, i, toFile)
+
           copyFile(path, toFile)
         end
       end if cs.include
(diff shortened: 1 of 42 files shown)


jerryk55:
Successfully fixed 95 of 600+ issues.  @roliveri @hsong-rh please review.  I'd like to have this merged and go on to more difficult issues as well.

lib/fs/MiqFsUtil.rb
```diff
@@ -251,28 +232,10 @@ def doUpdate(cs)
...
-            if !cs.nocompress || !cs.nocompress.detect { |ne| path =~ ne }
-              compressFile(path, toFile)
-              next
-            end
-          end
+
+          check_encrypt(cs, i, toFile)
+          check_compress(cs, i, toFile)
+
           copyFile(path, toFile)
```
roliveri: `copyFile` is now called unconditionally. Before this change, it was bypassed via the `next` calls above. Now that the above code has been moved into methods, the `next` calls will no longer work - they are probably not valid in the new methods, as there is no longer a corresponding loop.

lib/fs/MiqFsUtil.rb
```diff
@@ -322,6 +285,34 @@ def doUpdate(cs)
...
+  def check_encrypt(cs, i, toFile)
+    #
+    # If the file path matches an encrypt RE and doesn't
+    # match a noencrypt RE, then encrypt the contents of
+    # the file before copying it to the collection.
+    #
+    if cs.encrypt && cs.encrypt.detect { |e| i =~ e }
+      if !cs.noencrypt || !cs.noencrypt.detect { |ne| i =~ ne }
+        compressFile(i, toFile)
+        next
```
roliveri: This `next` no longer has a corresponding loop.

lib/fs/MiqFsUtil.rb
```diff
@@ -322,6 +285,34 @@ def doUpdate(cs)
...
+  def check_compress(cs, i, toFile)
+    #
+    # If the file path matches an compress RE and doesn't
+    # match a nocompress RE, then compress the contents of
+    # the file before copying it to the collection.
+    #
+    if cs.compress && cs.compress.detect { |e| i =~ e }
+      if !cs.nocompress || !cs.nocompress.detect { |ne| i =~ ne }
+        compressFile(i, toFile)
+        next
```
roliveri: This `next` no longer has a corresponding loop.

lib/fs/ntfs/attrib_index_root.rb
```diff
@@ -91,15 +91,12 @@ def allocations=(indexAllocations)
       if @indexNodeHeader.hasChildren? && indexAllocations
         indexAllocations.each { |alloc| @indexAllocRuns << [alloc.header, alloc.data_run] }
       end
-      @indexAllocRuns
```
roliveri: We never check the return value of this method?

lib/fs/ntfs/attrib_index_root.rb
```diff
@@ -91,15 +91,12 @@ def allocations=(indexAllocations)
...
     end
 
     def bitmap=(bmp)
       if @indexNodeHeader.hasChildren?
         @bitmap = bmp.data.unpack("b#{bmp.length * 8}") unless bmp.nil?
       end
-
-      @bitmap
```
roliveri: We never check the return value of this method?

lib/metadata/util/win32/Win32System.rb
```diff
@@ -163,7 +163,7 @@ def initialize(_c, fs)
       # @debug_str = ''
 
       # Force memory cleanup
-      software_doc = nil; sys_doc = nil; GC.start
```
roliveri: Don't we want to set these to nil so GC will clean them up?

lib/fs/MiqFsUtil.rb
jerryk55: Yup.  Not really sure what I was thinking.  Let me revisit this.

lib/fs/MiqFsUtil.rb
jerryk55: Ditto.

lib/fs/MiqFsUtil.rb
jerryk55: Ditto

lib/fs/ntfs/attrib_index_root.rb
jerryk55: Nor here.

lib/fs/ntfs/attrib_index_root.rb
jerryk55: And finally here - I think since the returned value is getting set to @bitmap it seemed to be overkill since its also getting set internally.  But that doesn't seem right either.  Time to revisit this as well.  Let me also find the CodeClimate complaint that had me address this and the previous one.

jerryk55: So the complaint by CodeClimate on the @bitmap and @indexAllocRuns issues is "Variable `@indexAllocRuns` used in void context."  I probably didn't totally understand the message - but it seems that since the variable is being set in the method, and then returned to set the same variable, there is some sort of overkill happening.  I apparently don't really understand this and should undo this two commits.

lib/metadata/util/win32/Win32System.rb
jerryk55: I see.  yes - CodeClimate (and I) didn't understand what was happening.  I will undo.

PR #66, opened 2018-05-14
Fix to remove extra leading spaces in /etc/fstab during SSA

hsong-rh:
Extra leading spaces in the entry of /etc/fstab will cause mount points missing. In our case, this cause packages information is empty after scanning finishes.

https://bugzilla.redhat.com/show_bug.cgi?id=1551273


hsong-rh:
@jerryk55 @roliveri This is the new one. Please use this to merge.

PR #67, opened 2018-05-14
Fix to change misleading error messages into normal info messages

hsong-rh:
https://bugzilla.redhat.com/show_bug.cgi?id=1450008


hsong-rh:
@roliveri @jerryk55 Please review

PR #70, opened 2018-05-16
Reduce MiqVm.rb Complexity

jerryk55:
CodeClimate complains that methods initialize and openDisks
in MiqVm.rb are too complex and too large.
In order to reduce both size and complexity, some of the functionality
was migrated to the relevant subclasses for Rhevm and Scvmm, where
appropriate.

@hsong-rh and @roliveri please review.  This is one that we've been talking about for a while - there was code in MiqVm that should have been moved into subclasses.  Since CodeClimate was complaining about the size and complexity of the methods this was an appropriate way to help.

Please note this moves the Maintainability score for this file from "C" to "A".

### DIFF
--- a/lib/MiqVm/MiqRhevmVm.rb
+++ b/lib/MiqVm/MiqRhevmVm.rb
@@ -18,10 +18,43 @@ class MiqRhevmVm < MiqVm
     rv
   end
 
+  def get_vmconfig(vm_config)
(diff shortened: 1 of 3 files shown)


jerryk55:
@roliveri this work was supposedly valid at the time.  I know we may not have a chance to review but is this something we should consider reopening, reviewing, and merging?

PR #71, opened 2018-05-17
Reduce Complexity of Win32Software.rb

jerryk55:
Reduce the complexity and size of the initialize and postProcessApps
methods by splitting them into component sections and other refactoring.
This along with other issues addressed by https://github.com/ManageIQ/manageiq-smartstate/pull/61 should reduce
the CodeClimate Maintainability score from D to A.

@roliveri @hsong-rh please review.  Continuing saga of addressing CodeClimate
complexity issues in this repo.

### DIFF
--- a/lib/metadata/util/win32/Win32Software.rb
+++ b/lib/metadata/util/win32/Win32Software.rb
@@ -98,7 +98,8 @@ module MiqWin32
 
     def registry_applications(registry_doc)
       # Get the applications
-      registry_applications_user_data(registry_doc)
+      reg_node = MIQRexml.findRegElement("HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Installer\\UserData", registry_doc.root)
+      registry_applications_user_data(reg_node) if reg_node
       ["HKEY_LOCAL_MACHINE\\SOFTWARE\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\App Paths",
        "HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths"].each do |reg_path|
         registry_applications_app_paths(registry_doc, reg_path)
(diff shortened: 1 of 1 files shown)


lib/metadata/util/win32/Win32Software.rb
```diff
@@ -80,160 +93,150 @@ def initialize(_c, fs)
...
+    def registry_applications_user_data(registry_doc)
+      reg_node = MIQRexml.findRegElement("HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Installer\\UserData", registry_doc.root)
+      reg_node&.each_element_with_attribute(:keyname) do |users|
+        users.each_element_with_attribute(:keyname) do |components|
+          next unless components.attributes[:keyname].downcase == "products"
+          components.each_element_with_attribute(:keyname) do |products|
+            attrs = XmlFind.decode(products, PRODUCTS_MAPPING)
+            attrs[:typename] = "win32_product"; attrs[:product_key] = @product_keys[attrs[:name]]
             clean_up_path(attrs)
-            @applications << attrs unless isDupApp?(attrs)
```
roliveri: @jerryk55 Why is the `unless isDupApp?(attrs)` no longer needed?

jerryk55: Good catch.  That got deleted accidentally.

jerryk55: Good catch.  Accidental deletion.

jerryk55:
@roliveri I've run the spec tests that exist, as well as running Smartstate on some VMs and checking that data has been filled in appropriately.  However I just noticed that I'd picked Linux VMs so let me try one for windows as well.  I'd see that a follow-on to some of this is also to improve test coverage for the repo as well.

jerryk55:
@roliveri and it took a while to get back down to acceptable within the rigourous CodeClimate restrictions but I was able to do so after refactoring a couple of more methods.  It is down to "A" score under Code Climate, Travis passes, and SSA on windows images runs successfully

PR #73, opened 2018-05-17
Fix to remove duplicated records in services list after SSA

hsong-rh:
The existing soft links in service config file cause the service report list has duplicate records. Skip these links when parsing them.

https://bugzilla.redhat.com/show_bug.cgi?id=1578792

### DIFF
--- a/lib/metadata/linux/LinuxSystemd.rb
+++ b/lib/metadata/linux/LinuxSystemd.rb
@@ -55,6 +55,11 @@ module MiqLinux
     def parse_service(file)
       debug "Parsing service unit: #{file}"
 
+      if @fs.fileSymLink?(file)
+        debug "#{file} is a soft link, skip to parse."
+        return nil
+      end
+
       unit        = @fs.fileBasename(file)
       name        = unit.gsub(".service", "")
       inif        = ini(file)
@@ -91,6 +96,11 @@ module MiqLinux
 
     def parse_target(file)
       debug "Parsing target unit: #{file}"
+
+      if @fs.fileSymLink?(file)
+        debug "#{file} is a soft link, skip to parse."
+        return nil
+      end
 
       unit = @fs.fileBasename(file)
       name = unit.gsub(".target", "")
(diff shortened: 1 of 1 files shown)


hsong-rh:
@roliveri @jerryk55 Please review.

jerryk55:
@hsong-rh is it possible to have the link in /etc/systemd/system that points to some .service file in some other directory besides  /usr/lib/systemd/system?  If that were the case then you would miss the service and you would count one less than you want.

hsong-rh:
@jerryk55 Ok, I will add check for this case, thanks.

hsong-rh:
@jerryk55 I added the check. @roliveri Please review.

lib/metadata/linux/LinuxSystemd.rb
```diff
@@ -119,6 +129,10 @@ def service_targets_xml(service)
       end
     end
 
+    def duplicated_link?(file)
+      existing_dirs.any? { |dir| @fs.getLinkPath(file).start_with?(dir) } if @fs.fileSymLink?(file)
```
roliveri: I'm not sure `start_with?` is good enough here, unless the values in `dir` end in a `/`. For example: `"/etc/dir11".start_with?("/etc/dir1")` is true.

hsong-rh: Good catch, I replaced `start_with?` with array comparison.

lib/metadata/linux/LinuxSystemd.rb
```diff
@@ -130,7 +130,11 @@ def service_targets_xml(service)
     end
 
     def duplicated_link?(file)
-      existing_dirs.any? { |dir| @fs.getLinkPath(file).start_with?(dir) } if @fs.fileSymLink?(file)
```
roliveri: I think it would be easier and faster just to change this to:
```ruby
existing_dirs.any? { |dir| @fs.getLinkPath(file).start_with?("#{dir}/") } if @fs.fileSymLink?(file)
```

hsong-rh: Changed.

PR #75, opened 2018-07-03
Fix to generate correct applications & patches info for windows scanning

hsong-rh:
Creating proper applications and patches list in scanning for window VMs.

https://bugzilla.redhat.com/show_bug.cgi?id=1597363

### DIFF
--- a/lib/metadata/util/win32/Win32Software.rb
+++ b/lib/metadata/util/win32/Win32Software.rb
@@ -52,12 +52,13 @@ module MiqWin32
     ]
 
     def initialize(_c, fs)
-      @patches = @applications = []
-      @patch_install_dates = @product_keys = {}
+      @patches = []
+      @applications = []
+      @patch_install_dates = {}
+      @product_keys = {}
 
       reg_doc = initialize_registry_doc(fs)
-
-      registry_applications(reg_doc)
+      registry_applications(reg_doc, fs)
       registry_patches(reg_doc)
     end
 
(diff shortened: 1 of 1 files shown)


lib/metadata/util/win32/Win32Software.rb
```diff
@@ -124,7 +125,7 @@ def registry_applications_user_data(reg_node)
       end
     end
 
-    def registry_applications_app_paths(registry_doc, registry_path)
+    def registry_applications_app_paths(registry_doc, registry_path, fs)
       return unless (reg_node = MIQRexml.findRegElement(registry_path, registry_doc.root))
       postProcessApps(reg_node, fs)
```
roliveri: @hsong-rh How did this ever work in the past? Do we know if there was a recent change that broke this?

lib/metadata/util/win32/Win32Software.rb
```diff
@@ -179,7 +180,7 @@ def registry_patches_packages(hotfix, element)
       return if element.attributes.nil? || element.attributes[:keyname].nil?
       if element.attributes[:keyname][0, 8] == 'Package_'
         # don't add this package if the ID is nil
-        return if (hotfix_id = hotfix_id(e)).nil?
```
roliveri: @hsong-rh Same here, was this refactored or something that caused the breakage?

lib/metadata/util/win32/Win32Software.rb
hsong-rh: Yes, it's broke by our recent refactor change https://github.com/ManageIQ/manageiq-smartstate/pull/71

lib/metadata/util/win32/Win32Software.rb
hsong-rh: Same change as above: https://github.com/ManageIQ/manageiq-smartstate/pull/71

PR #78, opened 2018-09-25
Parse 'docker ps' output

alexander-demichev:
Parse docker ps output if openstack was deployed with containerized services.

https://bugzilla.redhat.com/show_bug.cgi?id=1573507

### DIFF
--- a/lib/metadata/linux/LinuxUtils.rb
+++ b/lib/metadata/linux/LinuxUtils.rb
@@ -141,6 +141,27 @@ module MiqLinux
       end.compact
     end
 
+    def self.parse_docker_ps_list(lines)
+      return [] if lines.nil? || lines.empty?
+
+      lines.each_line.map do |line|
+        line = line.chomp
+        parts = line.split(/^(\S+)\s/)
+
+        name, = parts[1]
+
+        {:name              => name,
+         :systemd_load      => 'container',
+         :systemd_active    => parts[2].strip.start_with?('Up')? 'active' : 'failed',
+         :systemd_sub       => parts[2].strip.start_with?('Up')? 'running' : 'failed',
+         :typename          => 'docker_container',
+         :description       => parts[1] + ' ' + parts[2].strip,
+         :enable_run_level  => nil,
+         :disable_run_level => nil,
+         :running           => parts[2].strip.start_with?('Up')}
+      end.compact
+    end
+
     def self.collect_interface(interfaces, interface)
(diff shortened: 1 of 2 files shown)


lib/metadata/linux/LinuxUtils.rb
```diff
@@ -193,5 +214,52 @@ def self.parse_openstack_status(lines)
         }
       end
     end
+
+    def self.parse_openstack_container_status(lines)
```
mansam: As Rubocop mentioned, this method is pretty hard to read with all the chained blocks and conditionals. It should be broken down so that it's easier to follow.

spec/metadata/linux/LinuxUtils_spec.rb
```diff
@@ -64,4 +64,34 @@
       end
     end
   end
+
```
mansam: It would be a good idea to have tests here that expect the exact output of `merge_openstack_services` and `parse_openstack_container_status`.

lib/metadata/linux/LinuxUtils.rb
```diff
@@ -193,5 +214,52 @@ def self.parse_openstack_status(lines)
...
+              'name'    => service[1],
+              'active'  => service[2].strip.start_with?('Up'),
+              'enabled' => true
+            }
+          end
+        }
+      end
+    end
+
+    def self.merge_openstack_services(systemctl_services, containerized_services)
```
mansam: Probably a good idea to have some explanatory comments in here, or at least an introductory comment that explains what this does in a sentence.

alexander-demichev:
@agrare Hi, can you a look to this?

agrare:
@miq-bot assign @roliveri

lib/metadata/linux/LinuxUtils.rb
```diff
@@ -141,6 +141,27 @@ def self.parse_systemctl_list(lines)
...
+
+      lines.each_line.map do |line|
+        line = line.chomp
+        parts = line.split(/^(\S+)\s/)
+
+        name, = parts[1]
+
+        {:name              => name,
+         :systemd_load      => 'container',
+         :systemd_active    => parts[2].strip.start_with?('Up') ? 'active' : 'failed',
```
agrare: Maybe `&.strip&.start_with?` in case parts isn't what you're expecting?

alexander-demichev:
@roliveri Hi, can you review? :)

PR #80, opened 2018-10-26
Switch to manageiq-password from gems-pending

agrare:
Switch to the extracted manageiq-password gem from gems-pending.


agrare:
@miq-bot assign @roliveri

PR #81, opened 2018-11-08
Add ManageIQ::Password to the smartstate gemspec

agrare:
Add manageiq-password to the smartstate gemspec to fix issues loading the gem in dependent services


agrare:
@miq-bot assign @roliveri

PR #97, opened 2019-07-08
Add a rubocop channel to the codeclimate config

djberg96:
Add a rubocop channel to the codeclimate config

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -4,19 +4,17 @@ exclude_paths:
 - "**.xml"
 - "**.yaml"
 - "**.yml"
-- "lib/metadata/linux/test/Packages"
-- "lib/metadata/linux/test/tc_LinuxUtils.rb"
-- "locale/"
-- "spec/"
-- "test/"
-- "tools/"
-- "tmp/"
+- lib/metadata/linux/test/Packages
+- lib/metadata/linux/test/tc_LinuxUtils.rb
+- locale/
+- spec/
+- test/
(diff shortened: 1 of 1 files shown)


djberg96:
You can ignore the failure, this build is actually green: https://codeclimate.com/github/ManageIQ/manageiq-smartstate/builds/720

djberg96:
@miq-bot add_reviewer @roliveri

.codeclimate.yml
```diff
@@ -27,29 +25,28 @@ engines:
...
-    # let's enable later
     enabled: false
   markdownlint:
-    # let's enable later
     enabled: false
   rubocop:
     enabled: true
-    config: '.rubocop_cc.yml'
+    config: ".rubocop_cc.yml"
+    channel: rubocop-0-69
```
roliveri: @djberg96 Just making sure... Is this the only non-cleanup change to the file?

djberg96: Yeah, the rest is just the format that `YAML.dump` gives us.

roliveri: OK, good. Are the codeclimate failures the expected failures we discussed yesterday? If so, I'll merge it and cut a new version of the GEM.

djberg96: Yes. Not sure why you need to cut a new version of the gem, though, since it only affects our repo.

roliveri:
That's true. OK, no new GEM version needed.

PR #98, opened 2019-07-28
Remove v3 api form SSA

borod108:
Remove the usage of the ovirt gem and version 3 api from the SSA

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1535782

Depends on: https://github.com/ManageIQ/manageiq-providers-ovirt/pull/394

### DIFF
--- a/lib/MiqVm/MiqRhevmVm.rb
+++ b/lib/MiqVm/MiqRhevmVm.rb
@@ -25,11 +25,11 @@ class MiqRhevmVm < MiqVm
   end
 
   def getCfg(_snap = nil)
-    cfg_props = @rhevmVm.attributes
+    cfg_props = @rhevmVm.as_json.with_indifferent_access
 
     raise MiqException::MiqVimError, "Failed to retrieve configuration information for VM" if cfg_props.nil?
 
-    storage_domains = @rhevm.storagedomains
+    storage_domains = @rhevm.collect_storagedomains
     $log.debug "MiqRhevmVm#getCfg: storage_domains = #{storage_domains.inspect}"
 
     cfg_hash = {}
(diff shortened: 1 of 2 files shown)


borod108:
@roliveri is this ok?

borod108:
@miq-bot add_label bug
@miq-bot add_label ivanchuk/yes

agrare:
@miq-bot assign @roliveri

jerryk55:
I'm not sure this was looked at with regard to other interactions inside the SSA gem.  For instance, rhevmVm.disks was removed but it is used in the VmConfig code.

roliveri:
@borod108,
as per Jerry's comment, are there still issues with the `VmConfig` code?

borod108:
@jerryk55 can you suggest how I will find all those places? Currently with this code smartstate analysis works, how should I find all the places to change?

agrare:
@borod108 could you maintain the same interface but use the v3 api instead?  That should fix the issue of having to find every caller

borod108:
@agrare I could, we talked about it with Martin - it just does not seem to be a good solution. Also - if we fix it properly then the knowledge of how this thing actually works will spread to us as well and we will be able to fix things in the future and not keep putting bandages on this.

borod108:
@agrare @roliveri @jerryk55 so how should we proceed?
Can you point me to what else I should change?
I mean it runs on my env and there are no tests for this part that fail so I am not sure what to do here...

roliveri:
@jerryk55 Can you provide a pointer to where it's used in the `VmConfig` code?

jerryk55:
Yeah let me look it up again.  Hopefully I'm not mistaken.  Or maybe its better if I am.  More to come....

jerryk55:
I believe I was mistaken.  There is a reference to rhevmVm.disks in VMconfig - not rhevmVm.attributes[:disks] - I misread the code.  So I remove my objection to this.

roliveri:
@borod108 Will release a new version of the manageiq-smartstate GEM and let you know the version.

roliveri:
@borod108 released version 0.3.0 of the manageiq-smartstate GEM.

PR #100, opened 2019-09-02
Remove using ovirt gem

borod108:
There were still places using the ovirt gem that uses the V3 rhv api.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1535782

### DIFF
--- a/lib/metadata/VmConfig/VmConfig.rb
+++ b/lib/metadata/VmConfig/VmConfig.rb
@@ -478,10 +478,15 @@ class VmConfig
           end
         elsif miqvm.rhevm
           # First add the VM's active disk files.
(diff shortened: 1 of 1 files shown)


borod108:
@roliveri this is a blocker for the GA tomorrow, if we could get this merged it would be great.
@jerryk55 you were right, found those places in VmConfig :)

roliveri:
@borod108 I'll post the new GEM version here once it's done.

roliveri:
@borod108 New smartstate version = "0.3.1"

borod108:
@roliveri thank you!
https://github.com/ManageIQ/manageiq/pull/19245

PR #104, opened 2019-11-12
Don't include 'test' subdirectories in gem

simaishi:
Brought up during appliance build cleanup. I assumed all "test" subdirectories can be excluded from gem.

gem contents diff before/after change:

Only in before/lib/MiqVm: test
Only in before/lib/Scvmm: test
Only in before/lib: VmLocalDiskAccess  <-- this directory contains "test" subdirectory only
Only in before/lib/VolumeManager: test
Only in before/lib/fs/MetakitFS: test
Only in before/lib/fs/ext3: test
Only in before/lib/fs: test
Only in before/lib/metadata/MIQExtract: test
Only in before/lib/metadata/VmConfig: test
Only in before/lib/metadata/linux: test  <-- this directory contains 30MB `Packages` file
Files before/manageiq-smartstate.gemspec and after/manageiq-smartstate.gemspec differ

cc @bdunne @Fryguy @roliveri

### DIFF
--- a/manageiq-smartstate.gemspec
+++ b/manageiq-smartstate.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |spec|
   spec.license       = "Apache-2.0"
 
   spec.files         = `git ls-files -z`.split("\x0").reject do |f|
-    f.match(%r{^(test|spec|features)/})
+    f.match(%r{^(spec|features)/|test/})
   end
   spec.bindir        = "exe"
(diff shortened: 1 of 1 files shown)


Fryguy:
https://gist.github.com/simaishi/e49954866b8d689124c241586d09d819

Fryguy:
LGTM... @agrare can you also eyeball this?

PR #101, opened 2019-09-19
Add buildtime and url fields to collection from rpm database.

jeffmcutter:

PR #105, opened 2019-11-22
Update OpenStack image_from_snapshot error handling

aufi:
An ensure block in create_image_from_snapshot method with volume.destroy call can fail
for several reasons and we need find out why.

Added rescue block with more detailed logging output.

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1764847

### DIFF
--- a/lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
+++ b/lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
@@ -63,22 +63,28 @@ module MiqOpenStackCommon
     }
     volume = volume_service.volumes.new(volume_options)
     volume.save
-    begin
-      while volume.status != 'available'
-        sleep(10)
-        volume = volume_service.volumes.get(volume.id)
-      end
-      response = volume_service.action(volume.id, 'os-volume_upload_image' => {
-                                         :image_name => "Temp Image from #{snapshot.name}",
-                                         :disk_format => disk_format})
-      image_id = response.body["os-volume_upload_image"]["image_id"]
-      while image_service.images.get(image_id).status.downcase != 'active'
-        sleep(10)
-      end
-      return image_service.images.get(image_id)
-    ensure
-      volume.destroy
+
+    while volume.status != 'available'
+      sleep(10)
+      volume = volume_service.volumes.get(volume.id)
     end
+
+    response = volume_service.action(volume.id, 'os-volume_upload_image' => {
+                                       :image_name => "Temp Image from #{snapshot.name}",
+                                       :disk_format => disk_format})
+    image_id = response.body["os-volume_upload_image"]["image_id"]
+
+    while image_service.images.get(image_id).status.downcase != 'active'
+      sleep(10)
+    end
+
+    $log.debug "#{log_prefix}: Deleting temp volume #{volume.name} in #{volume.status} status"
+    volume.destroy
+
+    return image_service.images.get(image_id)
+  rescue => ex
+    $log.error "#{log_prefix}: Create image from Snapshot step raised error"
+    $log.error ex
   end
   
   def get_image_file_glance_v2(image_id)


lib/OpenStackExtract/MiqOpenStackVm/MiqOpenStackCommon.rb
```diff
@@ -63,22 +63,28 @@ def create_image_from_snapshot(snapshot_id, disk_format)
...
+                                       :image_name => "Temp Image from #{snapshot.name}",
+                                       :disk_format => disk_format})
+    image_id = response.body["os-volume_upload_image"]["image_id"]
+
+    while image_service.images.get(image_id).status.downcase != 'active'
+      sleep(10)
     end
+
+    $log.debug "#{log_prefix}: Deleting temp volume #{volume.name} in #{volume.status} status"
+    volume.destroy
```
aufi: ```volume.destroy``` was removed form ```ensure``` block for few reasons:
* it can (easily) fail too, since it relies on code few lines above and OSP APIs, and this failure is not handled
* if method failed before reaching volume.destroy line, most propably the volume even does not exist or attempt for its deletion would fail too
* currently, volume.destroy in ensure block doesn't fullfil its task to delete volume in all cases

aufi:
Rubocop issues mostly come from copied pieces of code already present in the source code file. I'm ready fix it in next step with potential fix of https://bugzilla.redhat.com/show_bug.cgi?id=1764847

jerryk55:
This looks good to me. However I don't have merge rights....  @roliveri ??

roliveri:
New manageiq-smartstate version: "0.3.3"

simaishi:
@roliveri The hammer build uses manageiq-smartstate 0.2.18.1, so we need this PR cherry-picked to 0.2.18.x branch and new gem released there.

Also, the new version 0.3.3 contains https://github.com/ManageIQ/manageiq-smartstate/pull/101 which requires a schema change (https://github.com/ManageIQ/manageiq-schema/pull/417) and this can't go to ivanchuk branch either. Please create a new branch and release a version that's < 0.3.3.

cc @Fryguy

roliveri:
@simaishi @Fryguy 

I have 0.2.18.x checked out and have cherry picked and committed the change.
I've updated the version file and staged it.
Can I now just to a rake release from that branch?

roliveri:
Also, Should I just update the version file on master to 0.4.0 and rake release. Then just delete the 0.3.3 branch?

simaishi:
> I have 0.2.18.x checked out and have cherry picked and committed the change.

Next time you cherry-pick, can you run the command using the merge commit, rather than a commit in the PR, using `git cherry-pick -x -m 1 <mergi commit SHA>` ? This way, we can clearly see which PR was cherry-picked.

> I've updated the version file and staged it.
> Can I now just to a rake release from that branch?

I think so, but I don't usually release gem, so...

roliveri:
Cut GEM version 0.2.18.2 with this change.
Cut GEM version 0.4.0 containing this change and https://github.com/ManageIQ/manageiq-smartstate/pull/101

simaishi:
What will happen to 0.3.x for ivanchuk without https://github.com/ManageIQ/manageiq-smartstate/pull/101?

roliveri:
@simaishi So, should https://github.com/ManageIQ/manageiq-smartstate/pull/101 be removed from 0.3.x and cur a new gem based on that (what version)? 
Or, should this change be cherry picked to 0.3.2 then cut the GEM version 0.3.2.1?

simaishi:
0.3.x makes more sense to me, but not sure if that's the right way to do... @Fryguy?

Fryguy:
According to SemVer...

0.4.0 is correct for the mainline gem, because it introduces forward compatible functionality.

0.2.18.2 is weird...it should have been 0.2.19 since it's a bug fix, but it's already out so it's fine now, and there was already an 0.2.18.1 for some reason.
 
> should this change be cherry picked to 0.3.2 then cut the GEM version 0.3.2.1?

Almost...this change should be cherry-picked on top of 0.3.2 and released as 0.3.3.  If we don't have an 0.3 branch, we can create one

simaishi:
I think 0.2.18.2 was done because only after 0.2.19 was released we realized it includes changes that can't be backported.

And we're facing the same issue here. 0.3.3 is already released with https://github.com/ManageIQ/manageiq-smartstate/pull/101, but that can't be used in ivanchuk branch.

Fryguy:
I was going based on this https://github.com/ManageIQ/manageiq-smartstate/pull/105#issuecomment-562698529 ... I didn't see that an 0.3.3 was released

Fryguy:
Then, yeah, we have to do an 0.3.2.1 or just cut 0.3.4 and remove the #101 commit somehow.

PR #108, opened 2020-01-21
Use rubygems.manageiq.org for handsoap

NickLaMuro:
Just mimicking what is being used in the `ManageIQ/manageiq` `Gemfile` currently:

https://github.com/ManageIQ/manageiq/blob/22858673dfae6b0018f3dc45aab8d94f7324a0ee/Gemfile#L16

Possibly we could just add an additional source at the top:

```ruby
source "http://rubygems.manageiq.org"
```

but just going to stay consistent for the time being with what we have in core for now.


Fryguy:
cc @bdunne

Fryguy:
This is green, but travis is not updating the PR, so I'm merging anyway.

PR #112, opened 2020-02-24
No need for MiqFaultTolerantVim without DRb

agrare:
Now that VMware connections are brokered by the OperationsWorker instead
of going over DRb to the VimBrokerWorker we no longer need
MiqFaultTolerantVim and can just use MiqVim.

https://github.com/ManageIQ/manageiq-providers-vmware/issues/520


lib/metadata/VmConfig/VmConfig.rb
```diff
@@ -412,7 +412,7 @@ def connect_to_host_vim(conn_reason, vmCfgFile)
       require 'VMwareWebService/miq_fault_tolerant_vim'
```
chessbyte: should this be removed as well?

agrare: It should be `/MiqVim`, updated

PR #114, opened 2020-02-26
Add Expanded Glob Processing for SSA

jerryk55:
Add find methodology from the virtfs gem.
find_class_methods.rb has some of the code from the same file in virtfs, as well as from v_dir.rb
in that gem.
Modify glob parsing in MD5deep to use this new FindClassMethods glob method instead.

Currently the XML processing aborts sometime after about 950 file entries.  This may be an already existing bug that is exacerbated by the amount of data now available when multiple glob characters can be entered in a scan profile.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1684589  (RFE).

@roliveri now available for initial review.

### DIFF
--- a/lib/metadata/util/md5deep.rb
+++ b/lib/metadata/util/md5deep.rb
@@ -60,12 +60,12 @@ class MD5deep
   def scan_glob(filename)
     filename.tr!("\\", "/")
     startDir = File.dirname(filename)
-    globPattern = File.basename(filename)
     @xml.root.add_attribute("base_path", startDir)
 
     # First check if we are passed a fully qualifed file name
     if @fs.fileExists?(filename)
-      isDir?(filename) ? process_dir_as_file(startDir, globPattern, @xml.root) : processFile(startDir, globPattern, @xml.root)
+      base_file = File.basename(filename)
+      isDir?(filename) ? process_dir_as_file(startDir, base_file, @xml.root) : processFile(startDir, base_file, @xml.root)
     else
       # If the file is not found then process the data as a glob pattern.
       begin
--- a/lib/metadata/util/md5deep.rb
+++ b/lib/metadata/util/md5deep.rb
@@ -56,6 +56,8 @@ class MD5deep
     filename.tr!("\\", "/")
     startDir = File.dirname(filename)
     @xml.root.add_attribute("base_path", startDir)
+    path_prefix = startDir[0, 2]
+    @drive_letter = path_prefix.match?(/^\w\:/) ? path_prefix : ""
 
     # First check if we are passed a fully qualifed file name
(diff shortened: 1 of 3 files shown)


lib/metadata/util/md5deep.rb
```diff
@@ -47,30 +48,48 @@ def scan(path, rootID = "/")
...
+    rescue Exception => err
+      $log.info "MD5.scan_glob: Exception #{err} rescued"
+      $log.debug err.backtrace.join("\n")
+    end
   end
 
   def scan_glob(filename)
     filename.tr!("\\", "/")
     startDir = File.dirname(filename)
     globPattern = File.basename(filename)
```
chessbyte: Is this still used with the new approach?

jerryk55: @chessbyte if you mean the globPattern variable, yes.

chessbyte: @jerryk55 it seems to be only used in the true leg of `if @fs.fileExists?(filename)` implying that it is not a glob pattern inside, but, rather, something else.  Suggest the `File.basename(filename)` call be moved inside the if statement and assigned to a more descriptive variable name.

jerryk55: @chessbyte agreed.  processFile does not treat it as a glob pattern - its really the basename of the file.  I'll move and rename it.

jerryk55: @chessbyte agreed.  processFile does not treat it as a glob pattern - its really the basename of the file.  I'll move and rename it.

roliveri:
Can you adapt any of the tests from `find_spec.rb`, of the `virtfs` gem, to work in the `manageiq-smartstate` gem? The test cases apply, but I'm not sure how easy it will be to get them to work without the core `virtfs`.

jerryk55:
@roliveri I'll take a look when I'm done addressing CodeClimate and RuboCop complaints...

jerryk55:
@roliveri adding some of the tests from the VirtFS gem and a couple of my own.  I removed the WIP label - I think this is ready to go with your approval.  Note that I will be adding a quick ManageIQ core PR to change the full pathname that is stored in the DB to be correct.  Without totally restructuring the XML built in md5deep.rb, the "filesystem" element name stored in the XML may or may not contain a glob.  It is used in the core Filesystem class to build the full pathname, when in actuality the full pathname has always been stored in the XML already in the fqname element.  I'll post the PR number here after submitting.

jerryk55:
@roliveri Is there more you'd like me to do with this prior to merging?  Thanks.

roliveri:
Released version 0.5.5 of manageiq-smartstate.

PR #115, opened 2020-02-27
Support compression in VMware sparse disk

chessbyte:
Much of the comments and understanding in this PR derives from the [VMware Virtual Disk Format 5.0 Documentation](http://www.vmware.com/support/developer/vddk/vmdk_50_technote.pdf) and this [repo](https://github.com/libyal/libvmdk)

### DIFF
--- a/lib/disk/modules/VMWareSparseDisk.rb
+++ b/lib/disk/modules/VMWareSparseDisk.rb
@@ -200,11 +200,7 @@ module VMWareSparseDisk
     # the grain tables.
     #
     buf = @vmwareSparseDisk_file.read(BYTES_PER_SECTOR)
-    @grainDirectory = []
-
-    GDES_PER_GD.times do |index|
-      @grainDirectory[index] = buf[index, GDE_SIZE].unpack1('L')
-    end
+    @grainDirectory = GDES_PER_GD.times.map { |index| buf[index, GDE_SIZE].unpack1('L') }
 
     #
     # In a Hosted Sparse Extent, all the grain tables are created when the sparse extent is created,
@@ -320,10 +316,10 @@ module VMWareSparseDisk
     gte = getGTE(grain_number)
 
     #
+    # when GTE >  1 - all reads: from the sparse disk
     # when GTE is 0 - reads with no parent: return 0s
     #               - reads with    parent: read from parent
     # when GTE is 1 - all reads: return 0s
-    # when GTE >  1 - all reads: from the sparse disk
     #
     if gte > 1
       read_grain_from_disk(gte, offset, length)
--- a/lib/disk/modules/VMWareSparseDisk.rb
+++ b/lib/disk/modules/VMWareSparseDisk.rb
@@ -5,7 +5,7 @@ require 'memory_buffer'
 require 'zlib'
 
 module VMWareSparseDisk
-  SPARSE_MAGIC_NUMBER = 0x564d444b # 'V' 'M' 'D' 'K'
+  SPARSE_MAGIC_NUMBER = "VMDK".unpack1("L>") # 0x564d444b
 
   SPARSE_EXTENT_HEADER = BinaryStruct.new([
     # the magic number is used to verify the validity of each sparse extent when the extent is opened
@@ -15,8 +15,8 @@ module VMWareSparseDisk
 
     # SparseExtentHeader is stored on disk in little-endian byte order, so if you examine the first eight bytes of a VMDK file,
     # you see either:
-    #   'K' D' 'M' 'V' 0x01 0x00 0x00 0x00
-    #   'K' D' 'M' 'V' 0x02 0x00 0x00 0x00
+    #   'K' 'D' 'M' 'V' 0x01 0x00 0x00 0x00
+    #   'K' 'D' 'M' 'V' 0x02 0x00 0x00 0x00
 
     # flags contains the following bits of information.
     'L',  'flags',                # bit 0:  valid new line detection test
@@ -202,7 +202,7 @@ module VMWareSparseDisk
     buf = @vmwareSparseDisk_file.read(BYTES_PER_SECTOR)
     @grainDirectory = []
 
-    (0..(GDES_PER_GD - 1)).each do |index|
+    GDES_PER_GD.times do |index|
       @grainDirectory[index] = buf[index, GDE_SIZE].unpack1('L')
     end
 
@@ -263,11 +263,7 @@ module VMWareSparseDisk
   end
 
   def parseGrainTable(buffer)
-    gt = []
-    (0..(GTES_PER_GT - 1)).each do |index|
-      gt[index] = buffer[index, GTE_SIZE].unpack1('L')
-    end
-    gt
+    GTES_PER_GT.times.map { |index| buffer[index, GTE_SIZE].unpack1('L') }
   end
 
   def readGrainTable(gde)
@@ -303,8 +299,8 @@ module VMWareSparseDisk
   def read_compressed_grain(grain_offset)
     @vmwareSparseDisk_file.seek(grain_offset, IO::SEEK_SET)
     buffer = @vmwareSparseDisk_file.read(SIZEOF_SPARSE_EXTENT_COMPRESSED_GRAIN_HEADER)
-    compressed_grain_header = OpenStruct.new(SPARSE_EXTENT_COMPRESSED_GRAIN_HEADER.decode(buffer))
-    buffer = @vmwareSparseDisk_file.read(compressed_grain_header.size)
+    compressed_grain_header = SPARSE_EXTENT_COMPRESSED_GRAIN_HEADER.decode(buffer)
+    buffer = @vmwareSparseDisk_file.read(compressed_grain_header['size'])
     Zlib::Inflate.inflate(buffer)
   end
 
--- a/lib/disk/modules/VMWareSparseDisk.rb
+++ b/lib/disk/modules/VMWareSparseDisk.rb
@@ -2,49 +2,117 @@ require 'disk/modules/MiqLargeFile'
 require 'disk/modules/miq_disk_common'
 require 'binary_struct'
 require 'memory_buffer'
+require 'zlib'
 
 module VMWareSparseDisk
+  SPARSE_MAGIC_NUMBER = 0x564d444b # 'V' 'M' 'D' 'K'
+
   SPARSE_EXTENT_HEADER = BinaryStruct.new([
-    'L', 'magicNumber',
-    'L', 'version',
-    'L', 'flags',
-    'Q', 'capacity',
-    'Q', 'grainSize',
-    'Q', 'descriptorOffset',
-    'Q', 'descriptorSize',
-    'L', 'numGTEsPerGT',
-    'Q', 'rgdOffset',
-    'Q', 'gdOffset',
-    'Q', 'overHead',
-    'C', 'uncleanShutdown'
+    # the magic number is used to verify the validity of each sparse extent when the extent is opened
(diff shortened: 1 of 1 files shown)


lib/disk/modules/VMWareSparseDisk.rb
```diff
@@ -2,49 +2,117 @@
...
     #
-    # Seek to start of the grain directory.
+    # If the grain directory sector number value is GD_AT_END and the "use redundant grain table" flag is set,
+    #  then the SPARSE_EXTENT_HEADER is located in the footer.
+    # The footer is a secondary file header stored at offset -1024 relative from the end of the file (stream)
+    #  that contains the correct grain directory sector number value.
     #
-    @vmwareSparseDisk_file.seek(@sparseHeader.gdOffset * blockSize, IO::SEEK_SET)
+    if flag_set?(@sparseHeader.flags, FLAG_MASK_USE_REDUNDANT_GRAIN_TABLE) && (GD_AT_END == @sparseHeader.gdOffset)
+      @vmwareSparseDisk_file.seek(-2 * BYTES_PER_SECTOR, IO::SEEK_END)
```
agrare: > `.seek(-2 * BYTES_PER_SECTOR, IO::SEEK_END)`

I think this should be `.seek(-SIZEOF_SPARSE_EXTENT_HEADER)` since next you `.read(SIZEOF_SPARSE_EXTENT_HEADER)`, it just happens to be that the sparse extent header is 2 sectors

lib/disk/modules/VMWareSparseDisk.rb
```diff
@@ -2,49 +2,117 @@
 require 'disk/modules/miq_disk_common'
 require 'binary_struct'
 require 'memory_buffer'
+require 'zlib'
 
 module VMWareSparseDisk
+  SPARSE_MAGIC_NUMBER = 0x564d444b # 'V' 'M' 'D' 'K'
```
Fryguy: Alternately,


```ruby
SPARSE_MAGIC_NUMBER = "VMDK".unpack("L>")[0]
```

lib/disk/modules/VMWareSparseDisk.rb
```diff
@@ -2,49 +2,117 @@
...
-    'C', 'uncleanShutdown'
+    # the magic number is used to verify the validity of each sparse extent when the extent is opened
+    'L',  'magicNumber',          # should be SPARSE_MAGIC_NUMBER
+
+    'L',  'version',              # the version number can be 1 or 2
+
+    # SparseExtentHeader is stored on disk in little-endian byte order, so if you examine the first eight bytes of a VMDK file,
+    # you see either:
+    #   'K' D' 'M' 'V' 0x01 0x00 0x00 0x00
+    #   'K' D' 'M' 'V' 0x02 0x00 0x00 0x00
```
Fryguy: Missing quotes before the D

lib/disk/modules/VMWareSparseDisk.rb
```diff
@@ -139,16 +189,67 @@ def d_size
...
+    @vmwareSparseDisk_file.seek(@sparseHeader.gdOffset * blockSize, IO::SEEK_SET)
+
+    #
+    # Read the first grain directory entry to get the offset to the start of
+    # the grain tables.
+    #
+    buf = @vmwareSparseDisk_file.read(BYTES_PER_SECTOR)
+    @grainDirectory = []
+
+    (0..(GDES_PER_GD - 1)).each do |index|
```
Fryguy: Prefer `(0...GDES_PER_GD)`  (3 dots means exclusive end)

Alternately

```
GDES_PER_GD.times do |index|
```

lib/disk/modules/VMWareSparseDisk.rb
```diff
@@ -2,49 +2,117 @@
...
+
+    'L',  'version',              # the version number can be 1 or 2
+
+    # SparseExtentHeader is stored on disk in little-endian byte order, so if you examine the first eight bytes of a VMDK file,
+    # you see either:
+    #   'K' D' 'M' 'V' 0x01 0x00 0x00 0x00
+    #   'K' D' 'M' 'V' 0x02 0x00 0x00 0x00
+
+    # flags contains the following bits of information.
+    'L',  'flags',                # bit 0:  valid new line detection test
```
Fryguy: I think there's a way to do bit level here, so you can actually give each of these flags an individual name here, and then you don't need the method that maps to a hash.

lib/disk/modules/VMWareSparseDisk.rb
```diff
@@ -161,6 +262,24 @@ def allocGrain(gn)
     sector * blockSize
   end
 
+  def parseGrainTable(buffer)
+    gt = []
+    (0..(GTES_PER_GT - 1)).each do |index|
```
Fryguy: I think this whole method can be

```ruby
GTS_PER_GT.times.map { |index| buffer[index, GTE_SIZE].unpack1('L') }
```

lib/disk/modules/VMWareSparseDisk.rb
```diff
@@ -180,4 +299,42 @@ def findFreeSector
     raise "Disk full." if @freeSector * blockSize > @capacity
     @freeSector
   end
+
+  def read_compressed_grain(grain_offset)
+    @vmwareSparseDisk_file.seek(grain_offset, IO::SEEK_SET)
+    buffer = @vmwareSparseDisk_file.read(SIZEOF_SPARSE_EXTENT_COMPRESSED_GRAIN_HEADER)
+    compressed_grain_header = OpenStruct.new(SPARSE_EXTENT_COMPRESSED_GRAIN_HEADER.decode(buffer))
+    buffer = @vmwareSparseDisk_file.read(compressed_grain_header.size)
```
Fryguy: Don't think we need the OpenStruct overhead just to pull put a single value

lib/disk/modules/VMWareSparseDisk.rb
chessbyte: I like the `times` approach.  Will change to that.

lib/disk/modules/VMWareSparseDisk.rb
chessbyte: That looks much cleaner!  Maybe even `unpack1` instead of `unpack(‘L>‘)[0]`

lib/disk/modules/VMWareSparseDisk.rb
chessbyte: I can add the padding if you think it will improve readability.

lib/disk/modules/VMWareSparseDisk.rb
Fryguy: I had no idea `unpack1` was a thing but that is WAY nicer.... `"VMDK".unpack1("L>")`

lib/disk/modules/VMWareSparseDisk.rb
chessbyte: Done!

lib/disk/modules/VMWareSparseDisk.rb
chessbyte: Thanks for the suggestion.  Implemented.

lib/disk/modules/VMWareSparseDisk.rb
chessbyte: Also, I thought I explained it in the comments above (derived from documentation)

lib/disk/modules/VMWareSparseDisk.rb
Fryguy: I tried doing this, but all the ways to do it are kind of clunky, so probably ok to go with what's here...

So I thought you could do 

```ruby
'b', :flag1,
'b', :flag2,
'b', :flag3,
...
```

Unfortunately the way the `b` identifier works is that it consume 1 but then moves the iterator forward the rest of the byte.  So, the only way to read consecutive bits is to do something like...

```ruby
'b8', :flags
```

which then gives something like `{:flags => "01100100"}`.

---

I think BinaryStruct could be enhanced with some sort of bit decoding, but I'm not sure what a nice interface would look like offhand.  I'm guessing the caller would just want to deal with bits individually, but under the covers as we build the format string we'd have to understand consecutive `b` definitions and pack them into a single `b<N>`.

Fryguy: That being said, it *might* be easier to use the bit decoding and do something like this...

```ruby
# Instead of
'L',  'flags'

# use this
'b16' 'flags'

# and then immediately after decoding do:

sparseHeader["flags"] = sparseHeader["flags"].chars.map {|c| c == "1"}

# Then perhps something like...

sparseHeader["valid_new_line_detection_test"] = sparseHeader["flags"][0]
sparseHeader["redundant_grain_table_used"]    = sparseHeader["flags"][1]
sparseHeader["zero_grain_gte_used"]           = sparseHeader["flags"][2]
sparseHeader["grains_compressed"]             = sparseHeader["flags"][16]

File: lib/disk/modules/VMWareSparseDisk.rb [hunk 8]
```diff
@@ -139,16 +189,67 @@ def d_size
...
+    #
+    # Read the first grain directory entry to get the offset to the start of
+    # the grain tables.
+    #
+    buf = @vmwareSparseDisk_file.read(BYTES_PER_SECTOR)
+    @grainDirectory = []
+
+    GDES_PER_GD.times do |index|
+      @grainDirectory[index] = buf[index, GDE_SIZE].unpack1('L')
+    end
```
Fryguy: This can be shortened to 

```ruby
    @grainDirectory = GDES_PER_GD.times.map { |index| buf[index, GDE_SIZE].unpack1('L') }
```

Fryguy: You can also unpack it in a single shot, which might be more efficient...

```ruby
    @grainDirectory = buf.unpack("L{GDES_PER_GD}")
```

lib/disk/modules/VMWareSparseDisk.rb
```diff
@@ -180,4 +295,42 @@ def findFreeSector
...
+
+  def read_grain(grain_number, offset, length, parent_offset)
+    gte = getGTE(grain_number)
+
+    #
+    # when GTE is 0 - reads with no parent: return 0s
+    #               - reads with    parent: read from parent
+    # when GTE is 1 - all reads: return 0s
+    # when GTE >  1 - all reads: from the sparse disk
+    #
```
Fryguy: Can you order the comment in the same order as the if/else conditions...I was trying to compare the 2 and it was confusing

Either that or...

```ruby
    if gte > 1
      # all reads: from the sparse disk
      ...
```

lib/disk/modules/VMWareSparseDisk.rb
chessbyte: shouldn't this be `b32` since we need to access bit 0 as well as bit 16 and bit 17?

lib/disk/modules/VMWareSparseDisk.rb
chessbyte: re-ordered comments to better match the code

lib/disk/modules/VMWareSparseDisk.rb
chessbyte: since GDE_SIZE is a constant with a value of 4 (i.e. not a `long`), went with:
```ruby
@grainDirectory = GDES_PER_GD.times.map { |index| buf[index, GDE_SIZE].unpack1('L') }
```

lib/disk/modules/VMWareSparseDisk.rb
Fryguy: Oh yes...I typoed it.

PR #117, opened 2020-03-24
Follow All Symlinks

jerryk55:
Currently SSA will follow symbolic links for every component in a path,
but if one component links to multiple files/directories only the first is followed.
This change will follow all the symbolic links for each component in a path.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1810194

@roliveri @hsong-rh please review.

### DIFF
--- a/lib/fs/modules/LinuxMount.rb
+++ b/lib/fs/modules/LinuxMount.rb
@@ -315,17 +315,24 @@ module LinuxMount
       #
       fs, lp = getFsPathBase(ncp)
       if fs.fileSymLink?(lp)
-        sl = getSymLink(fs, lp)
-        if sl[0, 1] == '/'
-          cp = sl
-        else
-          cp = File.join(cp, sl)
-        end
+        cp = follow_all_symlinks(fs, lp, cp)
       else
         cp = ncp
       end
     end
     (cp)
+  end
+
+  def follow_all_symlinks(fs, lp, cp)
+    while fs.fileSymLink?(lp)
+      sl = getSymLink(fs, lp)
+      if sl[0, 1] == '/'
+        lp = sl
+      else
+        lp = File.join(cp, sl)
+      end
+    end
+    lp
   end
 
   def getSymLink(fs, p)
--- a/lib/fs/modules/LinuxMount.rb
+++ b/lib/fs/modules/LinuxMount.rb
@@ -309,30 +309,27 @@ module LinuxMount
     #
     components.each do |c|
       ncp = File.join(cp, c)
-      #
-      # Each file system know how to check for,
-      # and read its own links.
-      #
-      fs, lp = getFsPathBase(ncp)
-      if fs.fileSymLink?(lp)
-        cp = follow_all_symlinks(fs, lp, cp)
-      else
-        cp = ncp
-      end
+      cp = follow_all_symlinks(ncp)
     end
     (cp)
   end
 
-  def follow_all_symlinks(fs, lp, cp)
-    while fs.fileSymLink?(lp)
-      sl = getSymLink(fs, lp)
-      if sl[0, 1] == '/'
-        lp = sl
+  def follow_all_symlinks(link_ptr)
+    #
+    # Each filesystem knows how to check for,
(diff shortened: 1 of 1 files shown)


lib/fs/modules/LinuxMount.rb
```diff
@@ -315,19 +315,26 @@ def expandLinks(p)
       #
       fs, lp = getFsPathBase(ncp)
       if fs.fileSymLink?(lp)
-        sl = getSymLink(fs, lp)
-        if sl[0, 1] == '/'
-          cp = sl
-        else
-          cp = File.join(cp, sl)
-        end
+        cp = follow_all_symlinks(fs, lp, cp)
```
roliveri: Every time we traverse to a new symlink, the filesystem, `fs`, can change.
I think `follow_all_symlinks` should be implemented at a higher level, so it can call `getFsPathBase` as needed.

jerryk55: Ah. Good catch @roliveri.  I'll look at that.

jerryk55:
@roliveri I've modified and re-tested with your suggestion.  Please review when possible.  Thanks.

lib/fs/modules/LinuxMount.rb
```diff
@@ -309,23 +309,27 @@ def expandLinks(p)
...
+    end
+    (cp)
+  end
+
+  def follow_all_symlinks(link_ptr)
+    #
+    # Each filesystem knows how to check for,
+    # and read its own links.
+    #
+    no_more_links = nil
```
roliveri: A nit, but can you set this to `true`?

lib/fs/modules/LinuxMount.rb
```diff
@@ -309,23 +309,27 @@ def expandLinks(p)
...
+    until no_more_links
+      filesys, link_ptr = getFsPathBase(link_ptr)
+      if filesys.fileSymLink?(link_ptr)
+        symlink = getSymLink(filesys, linkptr)
+        link_ptr = symlink[0, 1] == '/' ? symlink : File.join(comp, symlink)
```
roliveri: It appears `comp` isn't set.
I'm not sure what the value of `comp` should be when traversing nested, relative symlinks. That may require some investigation.

lib/fs/modules/LinuxMount.rb
```diff
@@ -309,23 +309,27 @@ def expandLinks(p)
...
-      if fs.fileSymLink?(lp)
-        sl = getSymLink(fs, lp)
-        if sl[0, 1] == '/'
-          cp = sl
-        else
-          cp = File.join(cp, sl)
-        end
+      cp = follow_all_symlinks(ncp)
+    end
+    (cp)
```
roliveri: I know this was like this before your change, but can you remove the parenthesis?

lib/fs/modules/LinuxMount.rb
jerryk55: If I set it to true then the loop will never get executed.  And the name of the variable will be incorrect.  Or are you suggesting I change the name of the variable and change the "until" to a "while"?

lib/fs/modules/LinuxMount.rb
jerryk55: Oh frig.  I copied the actual code into a different copy of the file from what I tested.  Hold on let me recommit this with the *correct* code!

lib/fs/modules/LinuxMount.rb
jerryk55: Yup.

lib/fs/modules/LinuxMount.rb
roliveri: Sorry, I meant `false` not `true`. I just meant to say use a boolean instead of `nil`.

jerryk55: Right.  Done.  :-)

jerryk55:
miq-bot add_label bug

roliveri:
Released version 0.5.4 of manageiq-smartstate.

PR #119, opened 2020-04-07
Fix Invalid Sort in Glob with Nil Array

jerryk55:
Fix check for valid return from check_file prior to
adding result to return array.
Failure to do so can result in an exception that will cause 
collected files not to be returned.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1821750

Related to https://bugzilla.redhat.com/show_bug.cgi?id=1684589

@roliveri please review.  Would like this ready for the upcoming build. Thanks.

### DIFF
--- a/lib/metadata/util/find_class_methods.rb
+++ b/lib/metadata/util/find_class_methods.rb
@@ -15,8 +15,9 @@ module FindClassMethods
 
     ra = []
     find(@search_path, glob_depth(glob)) do |p|
-      p = check_file(p, glob, flags)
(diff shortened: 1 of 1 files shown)


roliveri:
@jerryk55 do you want to try releasing the new version of the gem?

jerryk55:
@roliveri I will try but I have another one or two more PRs for SmartState I want to get into the build that I'll probably have done in the next day or two - does it make sense to wait so we're not doing a new version for each PR, or is that what we want?

PR #120, opened 2020-04-07
File collection exceptions

jerryk55:
If errors reading the disk to calculate the file digest are received,
information such as the file access time will be lost, and not written to
the xml file.  Make sure that calculate_digest handles the error to avoid this.
The upshot of an error like this is that an exception occurs parsing the
time data in the xml when it is supposed to be written to the database.
That exception would cause all scan_profile data to be lost.

One such disk/filesystem error that occurs is the following - 
The NTFS run spec LCN and length values need to be checked for nil values
prior to using them in the computation to determine which blocks to read.
If invalid data is read from the virtual disk, an exception will occur.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1821933

@roliveri please review.  I'd like to get this in next week's build if possible.

### DIFF
--- a/lib/metadata/util/md5deep.rb
+++ b/lib/metadata/util/md5deep.rb
@@ -232,9 +232,6 @@ class MD5deep
         while (buf = fileName.read(10_240_000))
           digest.each_pair { |_k, v| v << buf }
         end
-      rescue NoMethodError => method_err
-        $log.error "Exception NoMethodError #{method_err} reading file to calculate digest"
-        $log.debug method_err.backtrace.join("\n")
       rescue => err
         $log.error "Error #{err} reading file to calculate digest"
         $log.debug err.backtrace.join("\n")
--- a/lib/metadata/util/md5deep.rb
+++ b/lib/metadata/util/md5deep.rb
@@ -78,7 +78,7 @@ class MD5deep
       processFile(File.dirname(f), File.basename(f), @xml.root)
     end
   rescue => err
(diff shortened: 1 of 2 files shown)


lib/metadata/util/md5deep.rb
```diff
@@ -226,10 +226,18 @@ def calculate_digest(fileName)
...
-      # Loop over each digest and add the file contents
-      while buf = fileName.read(10240000)
-        digest.each_pair { |_k, v| v << buf }
+      begin
+        fileName.seek(0, IO::SEEK_SET)
+        # Loop over each digest and add the file contents
+        while (buf = fileName.read(10_240_000))
+          digest.each_pair { |_k, v| v << buf }
+        end
+      rescue NoMethodError => method_err
```
roliveri, line 235: Why do we get NoMehtodError? Wouldn't that mean there's a bug somewhere else in the code?

jerryk55: Well, yes.  And I did catch that bug with my change in fs/ntfs/data_run.rb.  I thought perhaps catching it even though I haven't seen it could address other issues that bad data on the disk could create.  I'd prefer to just catch the Exception class but our CodeClimate settings don't allow that.  If you prefer I can just leave in the "Rescue => err" clause, or I am open to other suggestions.

jerryk55: @roliveri removed the redundant rescue of NoMethodError as requested.  This should be ok to go with your approval.

PR #123, opened 2020-05-07
Fix Drive Letter Appending

jerryk55:
FindClassMethods.glob strips the Windows drive letter from files
it returns so we were previously prepending the letter back in
all cases in MD5deep::processFile.  However not all paths to processFile
go through glob, resulting in duplicate drive letters in some cases.
We will only prepend the letter after calling glob now.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1831645

@roliveri please review and merge when appropriate.  Thanks.

### DIFF
--- a/lib/metadata/util/md5deep.rb
+++ b/lib/metadata/util/md5deep.rb
@@ -72,9 +72,10 @@ class MD5deep
 
   def process_each_glob_file(file_name)
     FindClassMethods.glob(file_name, @fs) do |f|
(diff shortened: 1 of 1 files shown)


roliveri:
Is this the result of  a recent change? Why haven't we seen this before?

jerryk55:
Yup.  My changes for the glob processing did this.  Files that were returned by the glob processing were working fine but files that were found directly had this issue due to where I was appending the drive letter.

PR #124, opened 2020-05-08
Move miq_tempfile here

djberg96:
Let's move miq_tempfile.rb here, since smartstate is the only repo that uses it.

Part of the checklist for https://github.com/ManageIQ/manageiq-gems-pending/issues/231

I put it under `lib` since the gem configuration should make it available, but I can put it directly in the `OpenStackExtract/MiqOpenStackVm`directory if you prefer.


Fryguy:
@agrare Please review.

PR #125, opened 2020-05-11
Remove redundant lib directory

djberg96:
When moving stuff over from gems-pending this task got put under "lib", which resulted in a weird "lib/lib/tasks" layout. This PR just removes the redundant "lib".


djberg96:
@miq-bot add_reviewer @bdunne

djberg96:
@miq-bot add_reviewer @jerryk55

PR #129, opened 2020-05-14
Fix require for miq_tempfile

djberg96:
Missed this, not sure how the specs passed without it.

Followup to https://github.com/ManageIQ/manageiq-smartstate/pull/124


agrare:
@djberg96 maybe it was still pulling in the version in gems-pending

agrare:
Released v0.5.9

PR #133, opened 2020-07-16
More warning fixes

djberg96:
This is a followup to https://github.com/ManageIQ/manageiq-smartstate/pull/132. It addresses the last of the warnings the I see when running the smartstate specs locally.

* Fixes a redundant/wrong case statement in MiqSqlite3Cell.rb
* Fixes a circular require in MiqSqlite3Cell.rb
* ~~Comments out~~ Removes the inlined script in MiqVm.rb (which had a few unused variable warnings)
* Fixes a circular require in MiqOpenStackImage.rb
* Fixes a circular require in MiqOpenStackInstance.rb
* Fixes a circular require in MIQExtract.rb
* Fixes an unused variable warning in VmConfig.rb
* Initializes `@drive_letter` in md5deep.rb


djberg96:
@jerryk55, @roliveri How's it look?

lib/MiqVm/MiqVm.rb
```diff
@@ -345,3 +346,4 @@ def extract(c)
   vm.unmount
   puts "...done"
 end
+=end
```
roliveri, line 349: Most of this, in file  dev instance test code, hasn't been used in a very long time. If we think there will be no future need for it, it should probably be removed.

Otherwise, everything looks good to me.

djberg96: Alright, I'll update the PR and just yank it.

djberg96: @roliveri Updated.

PR #135, opened 2020-08-04
Bump vmware_web_service to 2.0

agrare:


agrare:
Thinking we should bump this to v0.6.0 so it wouldn't be pulled back to jansa

PR #136, opened 2020-08-11
s/to_miq_a/Array.wrap

d-m-u:
Fixing deprecation warnings and getting rid of the miq_a thing

@djberg96 here ya go


djberg96:
@d-m-u Unfortunately `Array.wrap` is a Rails thing, not a Ruby thing, so you'll need to add activesupport as a dependency as well.

PR #137, opened 2020-08-13
add presence check to array wrap to handle to handle blank case

d-m-u:
Fixing part of the bug I introduced in #136.

/review request @kbrock please

### DIFF
--- a/lib/metadata/MIQExtract/MIQExtract.rb
+++ b/lib/metadata/MIQExtract/MIQExtract.rb
@@ -126,7 +126,7 @@ class MIQExtract
 
   def extract(category, &blk)
     xml = nil
-    Array.wrap(category).each do |c|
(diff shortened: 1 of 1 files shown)


lib/metadata/MIQExtract/MIQExtract.rb
```diff
@@ -126,7 +126,7 @@ def getProfileData(&blk)
 
   def extract(category, &blk)
     xml = nil
-    Array.wrap(category).each do |c|
+    Array.wrap(category.presence).each do |c|
```
Fryguy, line 129: if category were nil this would have been fine previously.  In the PR title did you mean "to handle blank case"?

d-m-u: thanks

PR #139, opened 2020-09-18
RHEV SSA Fails on VMs with Direct LUNs

jerryk55:
There were assumptions in the code in three places that a RHEV VM's disk
storage_domains should exist when they do not for Direct Luns on a VM.
It looks like at attempt to rectify this was made previously by checking
if the ".first"  entry in the storage_domains array exists - but no attempt
was ever made to check if the array itself was nil.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1872954

@roliveri @djberg96 @Fryguy please review and merge as appropriate.  Thanks.


Fryguy (approved):
LGTM @agrare can you also review?

agrare (approved):

PR #145, opened 2020-10-21
Fix decode product key

Fryguy:
@chessbyte Please review.

I noticed this when going through the manageiq-gems-pending removals, and stumbled on https://github.com/ManageIQ/manageiq-gems-pending/pull/472.  However the search in that PR was incorrect as it wasn't searching the whole org.  When I searched the whole org I noticed these 2 callers of base24Decode.

This PR puts the base24_decode code back, and then fixes the callers which were actually broken, adding specs along the way.


lib/metadata/util/win32/Win32Software.rb
```diff
@@ -251,11 +251,13 @@ def convert_times(attrs)
...
-      return if product_key.blank? || product_key.length < 67
-      y = []; product_key.split(",")[52..67].each { |b| y << b.hex }
-      return MIQEncode.base24Decode(y)
+      parts = product_key.to_s.split(",")
+      return if parts.length < 67
+
+      ManageIQ::Smartstate::Util.base24_decode(parts[52..67].map(&:hex))
     rescue => err
       $log.error "MIQ(DecodeProductKey): [#{err}]"
+      nil
```
Fryguy, line 260: This is added because otherwise when a failure occurs the method returns `true` 😕 .  I couldn't figure out how to write a test for this though because I don't think it can fail 🤷‍♂️ .  I only noticed because I forgot to require manageiq/smartstate/util during my development and it was blowing up but my specs were returning `true` 😆

Fryguy:
We've likely not seen this in production because the rescue was catching the missing method and then just moving on.

lib/metadata/util/win32/Win32Software.rb
```diff
@@ -251,11 +251,13 @@ def convert_times(attrs)
     end
 
     def self.DecodeProductKey(product_key)
-      return if product_key.blank? || product_key.length < 67
-      y = []; product_key.split(",")[52..67].each { |b| y << b.hex }
-      return MIQEncode.base24Decode(y)
+      parts = product_key.to_s.split(",")
+      return if parts.length < 67
```
chessbyte, line 255: the code before was checking `product_key.length`, whereas the current code is checking `parts.length`.  Why the switch?

Fryguy: I had put it in the commit message, but forgot to put it in the OP

> The length check was incorrect as it checked the length of the string
> and not the length of the Array after splitting.

PR #149, opened 2021-01-20
Integrate miq-unicode.rb into smartstate

djberg96:
This moves the methods from miq-unicode.rb from gems-pending into the smartstate repo directly. Since we're reopening a base class (String), I thought this would be a good opportunity to use Ruby's refinements. This ensures that we don't cause any entanglements with any other libraries, now or in the future.

Part of the gems-pending effort: https://github.com/ManageIQ/manageiq-gems-pending/issues/231

Note that one other repo (virtfs-ntfs) uses a few of these methods so that will have to be addressed before we can remove miq-unicode.rb from gems-pending.

The virt_disk repo uses identical methods but defines them locally.

### DIFF
--- a/lib/miq_unicode.rb
+++ b/lib/miq_unicode.rb
@@ -1,43 +1,45 @@
 module ManageIQ
-  refine String do
-    def UnicodeToUtf8
-      dup.force_encoding("UTF-16LE").encode("UTF-8")
-    end
+  module UnicodeString
+    refine String do
+      def UnicodeToUtf8
+        dup.force_encoding("UTF-16LE").encode("UTF-8")
+      end
 
-    def UnicodeToUtf8!
-      force_encoding("UTF-16LE").encode!("UTF-8")
-    end
+      def UnicodeToUtf8!
+        force_encoding("UTF-16LE").encode!("UTF-8")
+      end
 
-    def Utf8ToUnicode
-      dup.force_encoding("UTF-8").encode("UTF-16LE")
-    end
+      def Utf8ToUnicode
+        dup.force_encoding("UTF-8").encode("UTF-16LE")
+      end
 
-    def Utf8ToUnicode!
-      force_encoding("UTF-8").encode!("UTF-16LE")
-    end
+      def Utf8ToUnicode!
+        force_encoding("UTF-8").encode!("UTF-16LE")
+      end
 
-    def AsciiToUtf8
-      dup.force_encoding("ISO-8859-1").encode("UTF-8")
-    end
+      def AsciiToUtf8
+        dup.force_encoding("ISO-8859-1").encode("UTF-8")
+      end
 
-    def AsciiToUtf8!
-      force_encoding("ISO-8859-1").encode!("UTF-8")
-    end
+      def AsciiToUtf8!
+        force_encoding("ISO-8859-1").encode!("UTF-8")
+      end
 
-    def Utf8ToAscii
-      dup.force_encoding("UTF-8").encode("ISO-8859-1")
-    end
+      def Utf8ToAscii
+        dup.force_encoding("UTF-8").encode("ISO-8859-1")
+      end
 
-    def Utf8ToAscii!
-      force_encoding("UTF-8").encode!("ISO-8859-1")
-    end
+      def Utf8ToAscii!
+        force_encoding("UTF-8").encode!("ISO-8859-1")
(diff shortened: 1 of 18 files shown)


lib/miq_unicode.rb
```diff
@@ -0,0 +1,43 @@
+module ManageIQ
```
Fryguy, line 1: Can this be a little more specific like `ManageIQ::UnicodeString`?  The `using ManageIQ` all over feels a bit scary as that's a top-level module

Fryguy:
Note that a lot of this code does not have automated tests.  Can you verify at least one of these calls is being executed through automated tests so we can know that everything works properly?

Fryguy:
cc @agrare

lib/miq_unicode.rb
djberg96: Ok, scoped it to `ManageIQ::UnicodeString` as per your suggestion.

djberg96:
> Note that a lot of this code does not have automated tests.

Like, *none* of it as far as I can tell. Oof.

> Can you verify at least one of these calls is being executed through automated tests so we can know that everything works properly?

I'll try to whip some specs for one of the files at least.

lib/miq_unicode.rb
```diff
@@ -0,0 +1,45 @@
...
+  module UnicodeString
+    refine String do
+      def UnicodeToUtf8
+        dup.force_encoding("UTF-16LE").encode("UTF-8")
```
chessbyte, line 5: should this be simplified to `dup. UnicodeToUtf8!`?

lib/miq_unicode.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+
+      def UnicodeToUtf8!
+        force_encoding("UTF-16LE").encode!("UTF-8")
+      end
+
+      def Utf8ToUnicode
+        dup.force_encoding("UTF-8").encode("UTF-16LE")
```
chessbyte, line 13: should this be simplified to `dup.Utf8ToUnicode!`?

lib/miq_unicode.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+
+      def Utf8ToUnicode!
+        force_encoding("UTF-8").encode!("UTF-16LE")
+      end
+
+      def AsciiToUtf8
+        dup.force_encoding("ISO-8859-1").encode("UTF-8")
```
chessbyte, line 21: should this be simplified to `dup.AsciiToUtf8!`?

lib/miq_unicode.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+
+      def AsciiToUtf8!
+        force_encoding("ISO-8859-1").encode!("UTF-8")
+      end
+
+      def Utf8ToAscii
+        dup.force_encoding("UTF-8").encode("ISO-8859-1")
```
chessbyte, line 29: should this be simplified to `dup.Utf8ToAscii!`?

lib/miq_unicode.rb
```diff
@@ -0,0 +1,45 @@
...
+      end
+
+      def Utf8ToAscii!
+        force_encoding("UTF-8").encode!("ISO-8859-1")
+      end
+
+      def Ucs2ToAscii
+        dup.force_encoding("UTF-16LE").encode("ISO-8859-1")
```
chessbyte, line 37: should this be simplified to `dup.Ucs2ToAscii!`?

lib/miq_unicode.rb
djberg96: Done.

lib/miq_unicode.rb
djberg96: Done.

lib/miq_unicode.rb
djberg96: Done.

lib/miq_unicode.rb
djberg96: Done.

lib/miq_unicode.rb
djberg96: Done.

djberg96:
I copied over the specs from gems-pending as well (and updated them slightly). They pass.

Fryguy (approved):
LGTM.  Will leave it to @chessbyte or @agrare

PR #150, opened 2021-01-21
Remove MetakitFS and related files

djberg96:
What started as an attempt to move miq-metakit.rb out of gems-pending resulted in the discovery that there is not, in fact, a metakit gem called "mk4rb" (or any metakit gem really). As per discussions on gitter, it turns out it was written about 14 years ago by a contractor and later dropped.

Which means that any code that actually attempts to use the thing will fail and should be removed. 

For this PR mostly that meant removing the MetakitFS directory and all of its files. Four other files referenced it: `VimDatastoreFS.rb` and `MiqMountManager.rb`, along with what appear to be a couple test scripts. I simply removed the test scripts. The relevant code in `VimDatastoreFS.rb` was commented out and harmless, but I updated it anyway.

The only code that needed modification was `MiqMountManager#findPayload`. Near as I can tell the method was useless without metakit, so I removed it. There was only 1 caller that I could find, also removed.


d-m-u:
do we want to do something about https://github.com/ManageIQ/virtfs/blob/acc63fef73b16edb4c5dea3d327eef88b3e76c0c/spec/kernel_spec.rb#L146 as well?

djberg96:
@d-m-u Yep, once this is removed, I'll update that one and gems-pending.

jrafanie:
@djberg96 can you search the org for any constants you're removing in this PR?

I did a [quick org search of one of them](https://github.com/search?q=org%3AManageIQ+VimDatastoreFS&type=code), `VimDatastoreFS`, and I believe this PR would break [this test code](https://github.com/ManageIQ/vmware_web_service/blob/7806fdadccb5cce7e6277d53ca85abd632ed8f08/test/VMwareWebService/browserTest.rb#L6)

In addition, we'll need to do a cross repo test in case we don't have tests that would expose other breakages.

Fryguy:
Old mk4rb source is here now, FWIW - https://github.com/ManageIQ/mk4rb

djberg96:
@jrafanie I searched the ManageIQ org, but I didn't see that one. Shouldn't matter though, since the code in `VimDatastoreFS` was commented out anyway.

jrafanie:
> @jrafanie I searched the ManageIQ org, but I didn't see that one. Shouldn't matter though, since the code in `VimDatastoreFS` was commented out anyway.

yeah, I think the require would fail

jrafanie:
> > @jrafanie I searched the ManageIQ org, but I didn't see that one. Shouldn't matter though, since the code in `VimDatastoreFS` was commented out anyway.
> 
> yeah, I think the require would fail

I was wrong.  I misread.  Sorry. It must have been all the deleted code, I got excited and saw deletions that weren't there.  🤣

Fryguy:
FYI, virtfs repos have all been archived now.

Fryguy:
@jrafanie is pleased

manageiq-smartstate 2021-01-25 12-43-51

Fryguy (approved):
LGTM - I searched around and can't find any callers.  I'm good with killing the lib/fs/test/copyTest.rb and lib/fs/test/updateTest.rb, because they appear to be metakit-specific tests

jrafanie:
How did I miss this?  

✂️ 🔥 ✂️ 🔥 ✂️ 🔥 

🍰  🍪 👏 🙇 😍 🎉

PR #152, opened 2021-01-25
Add explicit require statements for miq_unicode to modules that need them

djberg96:
Add explicit require's to modules that need them.

A couple other places where these methods are used are already picked up via MSCommon.rb.

Followup to https://github.com/ManageIQ/manageiq-smartstate/pull/149


Fryguy (commented):
Do you also have to add the `using` statement?

djberg96:
@Fryguy Yep, updated.

Fryguy (approved):

PR #153, opened 2021-01-27
Remove gems-pending runcmd references in favor of AwesomeSpawn

Fryguy:
This removes another dependency on gems-pending.

@chessbyte Please review.

For reference, this is the source of runcmd for comparison... https://github.com/ManageIQ/manageiq-gems-pending/blob/f984beaecc995a311d2f6e4b59cb57feb7003d75/lib/gems/pending/util/runcmd.rb#L3-L19

### DIFF
--- a/lib/metadata/VmConfig/xmlConfig.rb
+++ b/lib/metadata/VmConfig/xmlConfig.rb
@@ -1,5 +1,5 @@
 require 'util/miq-xml'
-require 'util/runcmd'
+require 'awesome_spawn'
 
 module XmlConfig
   def convert(filename)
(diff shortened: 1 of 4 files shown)


lib/metadata/VmConfig/xmlConfig.rb
```diff
@@ -12,9 +12,9 @@ def convert(filename)
       if Sys::Platform::IMPL == :linux
         begin
           # First check to see if the command is available
-          MiqUtil.runcmd("virsh list")
+          AwesomeSpawn.run!("virsh", :params => ["list"])
```
Fryguy, line 15: Note that unlike the other conversions the :combined_output, and .output are not needed here, since the output is ignored anyway.  All that is cared about here is whether the command blows up or not, which is covered by `run!`

chessbyte (approved):

PR #154, opened 2021-01-28
dos2unix all the files

djberg96:
@Fryguy noticed that somehow the fileformat on `directory_entry.rb` got set to dos instead of unix in https://github.com/ManageIQ/manageiq-smartstate/pull/152. I'm not sure what setting controlled this since I have the following in my .gvimrc:

```
set fileformat=unix
set fileformats=unix,dos
```

Anyway, I decided to run `find . -type f -print0 | xargs -0 dos2unix` and this updated a handful of files, including `directory_entry.rb`.


Fryguy (approved):

chessbyte:
why do the differences here on GitHub show a full delete and add?  I expected that [ignoring whitespace during diff](https://github.com/ManageIQ/manageiq-smartstate/pull/154/files?w=1) would show no changes, but 2 files diff strangely.

Fryguy:
I think part of the problem is that the dos line endings cause github to detect these files as "binary files", which doesn't play by the same "ignore whitespace" rules.  I manually verified these and they are the same.

PR #156, opened 2021-02-15
Add sqlite dev dependency

Fryguy:
This is used by spec/db/MiqBdb/data/conary/listpkg.rb

@agrare Please review.

Slightly related to https://github.com/ManageIQ/manageiq/pull/21052

### DIFF
--- a/manageiq-smartstate.gemspec
+++ b/manageiq-smartstate.gemspec
@@ -40,6 +40,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rake"
   spec.add_development_dependency "rspec", "~> 3.0"
   spec.add_development_dependency "simplecov"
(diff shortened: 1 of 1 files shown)


agrare:
I'm good with this but is this even used still?  I'm not even sure what a "Conary database" is

Fryguy:
Conary was an alternative package manager to RPM or dkpg.  It used to be on rpath.com, but that site seems dead now, so here's the closest information I could find - http://web.archive.org/web/20130118094547/http://wiki.rpath.com/wiki/Conary

I can't remember the details, but there were certain Linuxes (specifically [rPath Linux](https://distrowatch.com/table.php?distribution=rpath)) we had scanned that had conary db as the package manager, and smart state supported it for those Linuxes.

PR #181, opened 2024-02-13
Test with ruby 3.1 and 3.0

jrafanie:
Test with ruby 3.1 and 3.0


Fryguy:
@jrafanie What is this recorder change?  Is that a requirement to make it work with 3.1?

PR #189, opened 2024-06-11
Activate logical volume before SmartState scan if inactive

nasark:
If a logical volume is inactive, we should activate it before performing a smartstate scan

Depends on:
- [x] https://github.com/ManageIQ/manageiq-appliance/pull/387

@miq-bot assign @agrare 
@miq-bot add_reviewer @agrare, @Fryguy 
@miq-bot add_labels bug, radjabov/yes?

### DIFF
--- a/lib/MiqVm/MiqRhevmVm.rb
+++ b/lib/MiqVm/MiqRhevmVm.rb
@@ -1,3 +1,4 @@
+require 'awesome_spawn'
 require 'MiqVm/MiqVm'
 
 class MiqRhevmVm < MiqVm
@@ -51,12 +52,17 @@ class MiqRhevmVm < MiqVm
       storage_obj    = storage_domains_by_id[storage_id]
 
       file_path = file_path_for_storage_type(storage_obj, disk)
+      file_path_s = file_path.to_s
 
       tag = "scsi0:#{idx}"
       cfg_hash["#{tag}.present"]    = "true"
       cfg_hash["#{tag}.devicetype"] = "disk"
-      cfg_hash["#{tag}.filename"]   = file_path.to_s
+      cfg_hash["#{tag}.filename"]   = file_path_s
       cfg_hash["#{tag}.format"]     = disk.format
+
+      if storage_type_block?(storage_obj.storage.type) && !lv_active?(file_path_s)
+        AwesomeSpawn.run!("sudo lvchange", :params => [:activate, "y", file_path_s])
+      end
     end
     cfg_hash
   end
@@ -65,8 +71,7 @@ class MiqRhevmVm < MiqVm
     storage_type = storage_obj&.storage&.type
 
     # TODO: account for other storage types here.
-    case storage_type
-    when "nfs", "glusterfs"
+    if storage_type_file?(storage_type)
       add_fs_mount(storage_obj)
       fs_file_path(storage_obj, disk)
     else
@@ -176,5 +181,20 @@ class MiqRhevmVm < MiqVm
     rescue
       $log.warn "#{log_header} Failed to unmount all items from <#{nfs_mount_root}>.  Reason: <#{$!}>"
     end
+  end
+
+  private
+
+  # Output attributes of LV in column format and parse to retrieve active status
+  def lv_active?(lv_path)
+    AwesomeSpawn.run!("sudo lvs", :params => [:noheadings, :o, "lv_active", lv_path]).output.strip == "active"
+  end
+
+  def storage_type_block?(storage_type)
+    ["ISCSI", "FCP"].include?(storage_type.upcase)
+  end
+
+  def storage_type_file?(storage_type)
+    ["NFS", "GLUSTERFS"].include?(storage_type.upcase)
   end
 end


nasark:
Marking as WIP, its probably better to use awesomespawn as mentioned by @Fryguy

lib/MiqVm/MiqRhevmVm.rb
```diff
@@ -177,4 +181,11 @@ def unmount_storage
       $log.warn "#{log_header} Failed to unmount all items from <#{nfs_mount_root}>.  Reason: <#{$!}>"
     end
   end
+
+  private
+
+  # Output attributes of LV in column format and parse the 4th column to retrieve active status
+  def lv_active?(lv_path)
+    `sudo lvdisplay #{lv_path} -c | cut -d ":" -f4`.gsub("\n","") == "0" ? false : true
```
agrare, line 189: Let me check if there's a more programmatic way of checking this, but could you do `lvdisplay /dev/vg_system/lv_os | grep 'LV Status'|awk '{print $3}'`

e.g.
```
# lvdisplay /dev/vg_system/lv_os | grep 'LV Status'
  LV Status              available
# lvdisplay /dev/vg_system/lv_os | grep 'LV Status'|awk '{print $3}'
available
```

agrare: Per https://saratahosting.com/manpages/lvdisplay.8.html you can use `--colon` to print a more structured result which should be more reliable for machine parsing 
```
# lvdisplay --colon /dev/vg_system/lv_os 
  /dev/vg_system/lv_os:vg_system:3:1:-1:1:22020096:2688:-1:0:-1:253:0
```

Or lvs can take an explicit output column list:
```
# lvs -o lv_active /dev/vg_system/lv_os 
  Active
  active
```

I assume it is column title then value

nasark: On a test LV I created:
```
[manageiq@nasark vmdb]$ sudo lvdisplay /dev/vg_test/lv_test | grep 'LV Status'|awk '{print $3}'
available
[manageiq@nasark vmdb]$ sudo lvchange -an /dev/vg_test/lv_test
[manageiq@nasark vmdb]$ sudo lvdisplay /dev/vg_test/lv_test | grep 'LV Status'|awk '{print $3}'
NOT
```

nasark: When not active:
```
  --- Logical volume ---
  LV Path                /dev/vg_test/lv_test
  LV Name                lv_test
  VG Name                vg_test
  LV UUID               uuid
  LV Write Access        read/write
  LV Creation host, time , 2024-06-11 14:44:24 -0400
  LV Status              NOT available
  LV Size                800.00 MiB
  Current LE             50
  Segments               1
  Allocation             inherit
  Read ahead sectors     auto

```

Fryguy:
As discussed, I think it's better to use AwesomeSpawn here (if smartstate has access to AwesomeSpawn), otherwise use kernel spawn with proper args processing.

lib/MiqVm/MiqRhevmVm.rb
nasark: > Per https://saratahosting.com/manpages/lvdisplay.8.html you can use --colon to print a more structured result which should be more reliable for machine parsing

Yes the `-c` I'm using in the command is the same as `--colon`. The `lvs` option does look cleaner but returns empty when the lv is inactive. Let me know what you prefer
```
[root@nasark vmdb]# lvs --noheadings -o lv_active /dev/vg_test/lv_test 
        
```

lib/MiqVm/MiqRhevmVm.rb
```diff
@@ -57,6 +57,10 @@ def getCfg(_snap = nil)
       cfg_hash["#{tag}.devicetype"] = "disk"
       cfg_hash["#{tag}.filename"]   = file_path.to_s
       cfg_hash["#{tag}.format"]     = disk.format
+
+      if storage_obj.storage.type == "ISCSI" && !lv_active?(file_path.to_s)
+        `sudo lvchange -ay #{file_path.to_s}`
```
Fryguy, line 62: So, then, this would be something like

```suggestion
        AwesomeSpawn.run!("sudo lvchange", :params => [:a, :y, file_path.to_s])
```

I'm not 100% sure, but I think there's a better way to run via sudo as well - can't remember if kernel.spawn has that built-in or not.

lib/MiqVm/MiqRhevmVm.rb
nasark: with lvs we could do:
```
def lv_active?(lv_path)
    `sudo lvs --noheadings -o lv_active #{lv_path}`  == "active" ? true : false
end
```

agrare: Yeah I think the `lvs` way looks pretty clean

lib/MiqVm/MiqRhevmVm.rb
```diff
@@ -177,4 +182,11 @@ def unmount_storage
       $log.warn "#{log_header} Failed to unmount all items from <#{nfs_mount_root}>.  Reason: <#{$!}>"
     end
   end
+
+  private
+
+  # Output attributes of LV in column format and parse the 4th column to retrieve active status
```
Fryguy, line 188: Does this "4th column" comment still make sense?

lib/MiqVm/MiqRhevmVm.rb
```diff
@@ -57,6 +58,10 @@ def getCfg(_snap = nil)
       cfg_hash["#{tag}.devicetype"] = "disk"
       cfg_hash["#{tag}.filename"]   = file_path.to_s
       cfg_hash["#{tag}.format"]     = disk.format
+
+      if storage_obj.storage.type == "ISCSI" && !lv_active?(file_path.to_s)
```
Fryguy, line 62: Super-mega-minor, but every usage of file_path after line 54 calls .to_s, so maybe it's just better to call to_s once up there and then it doesn't have to be repeated here, nor below, nor on line 59.

lib/MiqVm/MiqRhevmVm.rb
```diff
@@ -51,12 +52,17 @@ def getCfg(_snap = nil)
...
+      file_path_s = file_path.to_s
 
       tag = "scsi0:#{idx}"
       cfg_hash["#{tag}.present"]    = "true"
       cfg_hash["#{tag}.devicetype"] = "disk"
-      cfg_hash["#{tag}.filename"]   = file_path.to_s
+      cfg_hash["#{tag}.filename"]   = file_path_s
       cfg_hash["#{tag}.format"]     = disk.format
+
+      if storage_obj.storage.type == "ISCSI" && !lv_active?(file_path_s)
```
agrare, line 63: I still think we need to account for other storage types like fibrechannel here, `file_path_for_storage_type` does this
```
    storage_type = storage_obj&.storage&.type

    # TODO: account for other storage types here.
    case storage_type
    when "nfs", "glusterfs"
      add_fs_mount(storage_obj)
      fs_file_path(storage_obj, disk)
    else
      lun_file_path(storage_obj, disk)
    end
```

Maybe we instead of checking storage.type == "something" we have a method like storage_type_file? and storage_type_block? to abstract this

lib/MiqVm/MiqRhevmVm.rb
agrare: I don't see this option in AwesomeSpawn or in Kernel.spawn's documentation

lib/MiqVm/MiqRhevmVm.rb
```diff
@@ -51,12 +52,17 @@ def getCfg(_snap = nil)
...
+        AwesomeSpawn.run!("sudo lvchange", :params => [:activate, "y", file_path_s])
```
agrare, line 64: @nasark do we need to de-activate the lv once SSA is complete?

nasark: I don't think that's the expectation

lib/MiqVm/MiqRhevmVm.rb
```diff
@@ -177,4 +182,19 @@ def unmount_storage
...
+
+  private
+
+  # Output attributes of LV in column format and parse to retrieve active status
+  def lv_active?(lv_path)
+    AwesomeSpawn.run!("sudo lvs", :params => [:noheadings, :o, "lv_active", lv_path]).output.strip == "active"
+  end
+
+  def storage_type_block?(storage_type)
+    ["ISCSI", "fibrechannel"].include?(storage_type)
```
agrare, line 194: I think it is "FCP" not "fibrechannel", https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.4/html/technical_reference/storage_domain_types1 

It _might_ be lowercase, in the Ovirt/RHV parsers we upcase it `storagedomain.dig(:storage, :type).upcase` and I see a bunch of places that .downcase before comparing.  I think the safest thing to do would be
```suggestion
    ["ISCSI", "FCP"].include?(storage_type.upcase)
```

And the same below with nfs / glusterfs

nasark: Thanks

agrare (approved):

PR #198, opened 2024-09-19
Fix MiqTempfile initialize kwargs error

agrare:
`Tempfile` takes a `**options` parameter which we were passing in as splat-ed positional parameters causing this error:

```
>> MiqTempfile.new("test", :encoding => 'ascii-8bit')
/usr/lib/ruby/3.1.0/tempfile.rb:134:in `initialize': wrong number of arguments (given 3, expected 0..2) (ArgumentError)
	from /home/grare/adam/src/manageiq/manageiq-smartstate/lib/miq_tempfile.rb:10:in `new'
	from /home/grare/adam/src/manageiq/manageiq-smartstate/lib/miq_tempfile.rb:10:in `initialize'
```

With this change applied:
```
>> MiqTempfile.new("test", :encoding => 'ascii-8bit')
=> #<File:/var/www/miq_tmp/test20240919-14037-7833j7>
```

https://github.com/ManageIQ/manageiq-providers-openstack/issues/890

### DIFF
--- a/lib/miq_tempfile.rb
+++ b/lib/miq_tempfile.rb
@@ -5,11 +5,11 @@ class MiqTempfile < DelegateClass(Tempfile)
   # TODO: share this definition with appliance console code.
   MIQ_TMP_DIR = '/var/www/miq_tmp'.freeze
 
-  def initialize(basename, *options)
+  def initialize(basename, **options)
     if File.directory?(MIQ_TMP_DIR)
-      super(Tempfile.new(basename, MIQ_TMP_DIR, *options))
+      super(Tempfile.new(basename, MIQ_TMP_DIR, **options))
     else
-      super(Tempfile.new(basename, *options))
+      super(Tempfile.new(basename, **options))
     end
   end
 end


Fryguy (approved):
Should we add specs too?

FWIW, CI only shows support for 3.0 and 3.1, so this should be fine.  Also this is only called from one place in our entire org and it's right in this gem.

agrare:
> Should we add specs too? (not sure it's worth it)

Yeah I started looking at this but I don't know that I could show the failure in a spec without actually calling Tempfile.new and hitting the filesystem

> Also this is only called from one place in our entire org and it's right in this gem.

Interesting maybe we could drop this and just use Tempfile normally in the caller

Fryguy:
> > Also this is only called from one place in our entire org and it's right in this gem.
> 
> Interesting maybe we could drop this and just use Tempfile normally in the caller

Yeah, I'm not sure.  I think the idea, based on the comment, was to have a single place to defined the /var/www/miq_tmp dir and be sure all callers are using it consistently, but the path is [all over the code base](https://github.com/search?q=org%3AManageIQ%20miq_tmp&type=code).

agrare:
Right the path is referenced in a number of places but Openstack SSA seems to be the only place actually using it, the rest of those are around configuring a disk (appliance_console) or monitoring the disk usage.

Based on the original commit message
```
    OpenStack fleecing: utilize extended temp storage.
    
    If additional temp storage has been added to the appliance - via appliance console -
    the OpenStack fleecing code will use it when downloading image contents.
```

It seems like the purpose of this class was to use this directory if it was configured, not just if it existed.  Maybe somewhere along the line we pre-created the directory even if a disk wasn't mounted?  Either way even back then it seems openstack ssa was the only caller of this class.

agrare:
It seems like on an appliance we will always have a /var/www/miq_tmp mountpoint https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/disk_layout.ks.erb#L7 so the "if /var/www/miq_tmp exists" doesn't make sense?

Fryguy:
> It seems like on an appliance we will always have a /var/www/miq_tmp mountpoint https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/disk_layout.ks.erb#L7 so the "if /var/www/miq_tmp exists" doesn't make sense?

I think the backup is more about non-appliance development and possibly just a safety net - better to have a backup than try to use a non-existent dir.

Fryguy:
Merging anyway cause this fixes a bug.  I am very curious how this is supposed to work on containerized though.

PR #201, opened 2024-10-31
EL9 RPM Sqlite3 DB support

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq-smartstate/pull/202
- [x] https://github.com/ManageIQ/manageiq-smartstate/pull/204
- [x] https://github.com/ManageIQ/manageiq-smartstate/pull/213


Fryguy:
I'd be happy merging specs as is ahead of actually implementing anything.

spec/metadata/linux/LinuxPackages_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+        expect(fs)
+          .to receive(:fileOpen)
+          .with(File.join(MiqLinux::Packages::RPM_DB, "Packages"), "r")
+          .and_return(File.open(File.expand_path('../../db/MiqBdb/data/rpm/Packages', __dir__), 'r'))
+      end
+
+      it "returns a list of rpm packages" do
+        result = described_class.new(fs)
+
+        expect(result.instance_variable_get(:@packages).count).to eq(690)
```
agrare, line 26: I'm not sure why we don't have an attr_reader for `@packages` here, `toXml` and `toString` are the only "interface" methods but I think access to packages would be handy also

lib/metadata/linux/MiqRpmPackages/Bdb.rb
```diff
@@ -0,0 +1,151 @@
...
+      1008 => "installtime",
+      1011 => "vendor",
+      1016 => "category",  # group
+      1020 => "url",
+      1022 => "arch",
+      1049 => "depends",   # requires
+    }
+
+    #
+    # Nubbers on disk are in network byte order.
```
Fryguy, line 51: LOL nubbers

lib/db/MiqSqlite/MiqSqlite3Table.rb
```diff
@@ -63,7 +63,7 @@ def decodeSchema
       return if @type != "table" || @name[0..6] == "sqlite_"  # Names beginning with sqlite_ are internal to engine
       @columns = []
       sql = @sql.gsub(/[\n\r]/, "")
-      re1 = /\s*CREATE\s+TABLE\s+(\w+)\s*\((.*)\)\s*/
+      re1 = /\s*CREATE\s+TABLE\s+'?(\w+)'?\s*\((.*)\)\s*/
```
Fryguy, line 66: That's interesting - I've never seen quotes before.  Also note that technically the table name could have a `.` in it as well (i.e. `<schema_name>.<table_name>`)

Fryguy:
@agrare More out of curiosity, what's the approach here?  I'm seeing refactoring into separate modules + specs...is the plan to do an initial refactor, then merge, then add the new DB support?

agrare:
@Fryguy so I see this as two high level issues,

1. MiqSqlite3DB can't currently load the rpmdb.sqlite database
2. MiqRpmPackages can't currently process a rpmdb.sqlite file


1. Get `MiqSqlite3DB::Sqlite3` able to load the rpmdb.sqlite file and list tables
3. Get `MiqSqlite3DB::Sqlite3Table` able to list rows in the Packages table (this is also broken currently)
4. Get `MiqLinux::MiqRpmPackages` able to understand the contents of the rpmdb.sqlite database (this PR)

So I'm rebasing this on top of the sqlite3 fixes

Fryguy:
Moving forward I wonder if it's better to use the sqlite library instead of rolling our own.  Not sure if that is with investigation or not.  I'd rather do whats more expedient at this time, but I'm also wondering what bugs will lurk in the corners.

agrare:
> Moving forward I wonder if it's better to use the sqlite library instead of rolling our own

I looked into that a bit, but I don't think that will work with the way that we handle loading the filesystem via MiqFs in SSA.  We'd need a "real" file since the interface is a path and all of the loading of the database is done in C-ext code.

lib/metadata/linux/MiqRpmPackages/Sqlite.rb
```diff
@@ -0,0 +1,55 @@
...
+    def initialize(fs, dbFile)
+      @fs           = fs
+      @db_file_path = dbFile
+
+      @rpmdb_tempdir = nil
+      @rpmdb_path    = nil
+
+      if fs.present?
+        @rpmdb_tempdir = Dir.mktmpdir("rpmdb-")
+        FileUtils.mkdir_p("#{@rpmdb_tempdir}/var/lib/rpm")
```
Fryguy, line 12: For Mac support, the path needs to be `"#{@rpmdb_tempdir}/opt/homebrew/var/lib/rpm"`

lib/metadata/linux/MiqRpmPackages/Sqlite.rb
```diff
@@ -0,0 +1,68 @@
...
+class MiqRpmPackages
+  class Sqlite < MiqRpmPackages
+    def initialize(fs, dbFile)
+      @fs           = fs
+      @db_file_path = dbFile
+
+      @rpmdb_tempdir = nil
+      @rpmdb_path    = nil
+
+      if fs.present?
```
Fryguy, line 13: I had to change the specs to stub this, but I think .present? from active_support is not available at this context, so we either need to require it, or implement differently.

agrare: I can replace with just `if fs`, other callers in smartstate just do `if @fs.nil?; else`

Fryguy:
Nice - tests went green.

agrare:
Okay checklist of things that we need for this to be merged:
- [ ] ruby-rpm-ffi support for rpm 4.14+ (mostly working in master, but no release cut) so that requiring the gem doesn't blow up with librpm.so.8+
- [ ] ruby-rpm-ffi add detection of librpm.so.9 and librpm.so.10 (we can work around this by installing rpm-devel so ruby-rpm-ffi picks up on librpm.so but that is sub-optimal)

spec/metadata/linux/data/rpm/rpmdb.sqlite
agrare, line 1: @Fryguy I took the rpmdb.sqlite from one of the ruby-rpm-ffi tests where it installs a single dummy package so that we can actually test the package parser code above instead of just `expect(result.packages.count).to eq(0)`

agrare:
I've updated this to use the forked gem, trying to figure out why this is red on CI but green locally

agrare:
Ah I figured it out, on the ubuntu runner (and Fedora 41) the `RPM["_dbpath"]` is `"/usr/lib/sysimage/rpm"` so our rpmdb.sqlite that we extracted to TEMPDIR/var/lib/rpm wasn't being found.

lib/metadata/linux/MiqRpmPackages/Sqlite.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    def close
+      FileUtils.rm_rf(@rpmdb_tempdir) if @rpmdb_tempdir
+    end
+
+    private
+
+    def rpm_db_relative
+      @rpm_db_relative ||= begin
+        parts = [RbConfig::CONFIG["host_os"] =~ /darwin/ ? "opt/homebrew" : nil, RPM['_dbpath']].compact
```
agrare, line 63: @Fryguy I hope we can use the _dbpath constant on OSX also and then we could do away with this method :crossed_fingers:

agrare:
Okay taking out of WIP as this is passing specs with a populated rpmdb.sqlite database, we can figure out the behavior of the _dbpath on mac now or in a follow-up

lib/metadata/linux/LinuxPackages.rb
```diff
@@ -155,9 +155,13 @@ def procPortage(pd)
     #
     def procRPM(dbDir)
       $log.debug "Processing RPM package database"
-      rpmp = MiqRpmPackages.new(@fs, File.join(dbDir, "Packages"))
-      rpmp.each { |p| @packages << p }
-      rpmp.close
+
+      rpmp = MiqRpmPackages.new(@fs, dbDir)
```
Fryguy, line 159: Eventually we should add a block form of this that auto-closes

agrare:
NOTE in order to support EL10 we'll need to check `/var/lib/rpm` as well as `/usr/lib/sysimage/rpm`

Fryguy:
> NOTE in order to support EL10 we'll need to check `/var/lib/rpm` as well as `/usr/lib/sysimage/rpm`

would that be covered by `_dbpath`?

lib/metadata/linux/MiqRpmPackages.rb
```diff
@@ -1,152 +1,29 @@
...
-
-  def getStringArray(data, offset, count)
-    ra = []
-    cpos = offset
-
-    count.times do
-      s = getStringVal(data, cpos)
-      ra << s
-      cpos += (s.length + 1)
+      orig_new(fs, dbDir)
```
Fryguy, line 25: I think we can use a `super` pattern here instead of the orig_new

agrare:
`RPM["_dbpath"]` covers where the local system thinks the db should be, when we look to the system that we're scanning the db path could be one or the other depending on the version of that system.

lib/metadata/linux/MiqRpmPackages.rb
Fryguy: Even so, I will merge without this.

agrare: I pulled this pattern from linux_admin, if this isn't the best way to do it we should probably change both then

Fryguy (approved):

PR #209, opened 2024-12-10
Fix codeclimate-action version check syntax error

agrare:


Fryguy:
Dang, I thought I found all of these :/

Fryguy:
Here's the master run after this PR merged: https://github.com/ManageIQ/manageiq-smartstate/actions/runs/12260717759

PR #210, opened 2024-12-10
Add a version limit on activesupport gem

agrare:
Without an upper limit on activesupport this was bringing in v8.0.0 which was conflicting with core's uri version.


jrafanie (approved):
LGTM for now until we move to the newer URI.  I'm hesitant to add another dependency we need to loosen later on but I see no other solutions for now.  It should be tested in a similar set of common dependencies as core.  It's not really a standalone gem.

